PR #3, opened 2018-01-05
DSL rewrite

solnic:
This is a complete rewrite of the DSL, originally ported from dry-validation. This is the first step towards dry-validation 1.0.0 which will be built on top of this gem.

### Goals

- much simpler implementation with a significant CLOC reduction and simplified way of testing
- support for chaining macros ie `required(:numbers).filled(:array?).each(:int?)`
- trivial way of adding your own macros

#### Phase 1

Basic implementation of the DSL and porting message APIs from dry-validation.

- [x] make `required` work
- [x] make `value` work
- [x] make `optional` work
- [x] make `each` work
- [x] restore support for defining predicates in blocks
- [x] support for chaining macros
- [x] make nested schemas work
- [x] restore message compiler
- [x] improve class organization and reduce duplication
- [x] fix an issue where nested schema is missing a name (causes lack of the key name in error output too)
- [x] restore type-spec API

#### Phase 2

Porting relevant specs from dry-validation.

- [x] remove `__lib` which I temporarily kept only as a reference during development
- [x] restore specs from dry-validation, excluding cases we no longer want to support (this will take a lot of time):
    - [x] port __spec/integration/schema/predicates/*
    - [x] port __spec/integration/schema/explicit_type_specs
    - [x] port __spec/integration/schema/macros/*
      - [x] port __spec/integration/schema/macros/value
      - [x] port __spec/integration/schema/macros/filled
      - [x] port __spec/integration/schema/macros/each
      - [x] port __spec/integration/schema/macros/maybe
    - [x] port __spec/integration/schema_spec
    - [x] port __spec/integration/form/predicates/*
    - [x] port __spec/integration/messages/*
    - [x] port __spec/integration/custom_error_messages_spec
    - [x] port __spec/integration/custom_predicates_spec
    - [x] port __spec/integration/hints_spec
    - [x] port __spec/integration/injecting_rules_spec
    - [x] port __spec/integration/localized_error_messages_spec
    - [x] port __spec/integration/optional_keys_spec
    - [x] port __spec/integration/result_spec

#### Phase 3

General clean-up / smaller refactorings / doc-writing / minor improvements.

- [x] Refactor macro class hierarchy (`Macros::Core` shouldn't refer to its descendants)
- [x] Simplify code handling nested schemas
- [x] ~~Fix failure on Travis~~ I marked it as pending for now, i18n fallback API changed, hence the failure
- [x] ~~Make `negation` pending specs pass~~ this needs some considerations, ie how do we handle error messages with negated predicates? in dry-v it is only supported by high-level rules, which we don't support here, so I left it as a pending spec


solnic:
Done with phase 1 :) I moved dry-v integration specs to `spec/__integration` and now we just need to make most of them pass. I'm gonna start with `spec/__integration/schema/predicate`.

hwo411:
I was playing a little with `form` specs and dig into `dry-validation` implementation which used input processor to coerce the type.

One of the examples I tried in `dry-validation` is
```ruby
Dry::Validation.Form do
  required(:q).maybe(:str?, :int?)
end
```

calling which results in

`
s.({ q: '1' })
 => #<Dry::Validation::Result output={:q=>1} errors={:q=>["must be a string"]}>
`

Which probably should be valid on forms, but due to coercion can't be. Also this and the current `dry-schema` implementation make me think that coercion is probably not the best option here, but it's needed for output, and it'd be better to change DSL to specify type explicitly to disallow any ambiguity and disable passing it together with other predicates.

I was thinking about extracting type as an optional separate call in the following way:

```ruby
required(:q).as(:int?).filled(gt?: 18)
```

Where `as` requires a type to be passed, which can be just a check and can be a check with coercion for the forms. And another idea(from C#) here is to allow passing both `:int` and `:int?` to `as`, where `int` is strictly integer and `int?` is nullable integer. To pass schema symbol starting with Capital letter can be used, e.g. `as(:Dog)`.

As an extension to this to support `anyOf`, `allOf`, `oneOf`, `not` etc. for schemas, `as` can be replaced with them, e.g.

```ruby
required(:q).any_of(
  as(:int).filled(gt?: 18),
  as(:string),
  as(:Car?)
)
```

WDYT?

volkanunsal:
@solnic Can you give us an example of how custom predicates would work in the new DSL?

solnic:
@hwo411 sorry for late reply, these are very interesting suggestions actually. I'll experiment with your ideas when working on this refactor and get back to you when I have something working.

> And another idea(from C#) here is to allow passing both :int and :int? to as, where int is strictly integer and int? is nullable integer

We may consider using such syntax, esp that it exists in other langs like you mentioned (TypeScript has it to, FWIW).

solnic:
> @volkanunsal Can you give us an example of how custom predicates would work in the new DSL?

I actually don't think dry-schema will support custom predicates, it's something dry-validation can provide instead. The reason is that dry-schema should be much much simpler and focused on strict type checking and coercion, whereas dry-validation would be a higher level abstraction with more complex validations, where custom predicates can be useful.

volkanunsal:
@solnic I can see that. However, type refinement is a pretty basic feature for type definition. Some other libraries I've worked with that do the same job have it. Case in point: [tcomb](https://github.com/gcanti/tcomb/). All it would be needed is to add a refinement API for dry-logic, and feed it to dry-schema:

```ruby
types = Dry::Logic::Predicates.add do |t|
  t.refine :email?, -> value { value.include?('@') }   
end

Dry::Shema.define(types) do
...
end
```

solnic:
Hey I'm back.

Got `spec/integration/schema/predicates` passing and started working on porting `spec/__integration/form/predicates`. Initial work to make form predicates pass required adding support for type specs known from dry-validation, and now they are *obligatory* if you want coercion to kick in. You can see the details in here https://github.com/dry-rb/dry-schema/pull/3/commits/c1e078aca45d0b658c541850052921c91449f0d0 - this is very basic, but it was enough to make `eql?` specs pass. I'll expand type specs functionality in another PR, here I just want specs to pass so that we can get this beast merged into master soon.

solnic:
I'm getting very close to finish this PR. See updated TODO list. I know there are many un-checked items, but those won't take too long. I probably need roughly 20 more hours to put into this. I should be able to finish this in the first week of October.

PR #11, opened 2018-11-17
Type specs with filter rules

solnic:
This adds the ability to pass type specs as the first argument in `#value` and `#filled` or via explicit `#type` method in DSL. It also introduces `#filter` method which can be used to define pre value coercion rules.

### API

``` ruby
Dry::Schema.define do
  required(:age).filter(format?: /\d+/).value(:int, gt?: 18)
  # or
  required(:age).filter(format?: /\d+/).type(:int).value(gt?: 18)
end
```

This will:

1) Check `format?`
2) Coerce to `:int`
3) Apply `value` rules

### Summary of changes

- A special `input_schema` was added to the DSL, it is used by `Key` macros (required/optional) to define filter rules using the new `#filter` method
- If `input_schema` has any rules, its definition will be added to the processor right after key coercion
- `Key` macro overrides `#value` and `#filled` to support type specs
- `Result` is now created by the processor and it's passed to its steps, if a step returns a hash back, result's output will be replaced by it. This is mostly a performance tweak so that we're not creating multiple results as it's an expensive object. After processing is done, a frozen result instance is returned.


Morozzzko:
One of the things that made it harder to persuade my colleagues to use dry-validation is the feeling of redundancy. Let me elaborate on that.

You see, in the former type-specs DSL we would write required(:age, :integer).value(:int?), which made people confused: why would we say that we want an integer two times? 

Now the DSL seems to be getting even more complicated: the line required(:age).type(:int).input(format?: /d+/).value(:int?, gt?: 18) specifies an integer *three* times! That's a lot! 

Perhaps, `:int?` is redundant. Why would you check an `:int` if it's an `int?`?

So, we could probably live with something like this:

```required(:age).input(format?: /d+/\).type(:int).value(gt?: 18)```

Alright, we've got rid of a redundant check. Now there's another problem for people: we have two methods that perform similar task — they specify the output type.

I see that an `:int` which is `greater than 18` is just a constrained / dependent type. We should use a single method for that. 

I don't know what name is better, but let's say we use `output` instead of `value` for the sake of semantics.

```ruby
required(:age).input(format?: /d+/\).output(:integer, gt?: 18)

required(:age).input(format?: /d+/\).output(Types::Strict::Integer.constrained(gt?: 18))
```

Alright. This is probably better, but now we've got another problem — we used to have three methods for extra checks:

* `value`
* `filled`
* `maybe`

The truth is: we don't need that much. It's confusing. Not to me — but other developers. I propose we get rid of them. 

Steps to do so:

* Replace `value` with `output`
* Replace `filled` with `output(:filled?)` ← going back to the origins, eh
* Replace `maybe` with `output(:integer, :nil)`, `output(:integer, :nilable)` or `output(:integer, :optional)`


---- 

TL;DR

Original dry-validation DSL is confusing to a lot of people which makes it hard to promote the schema.

Main points of confusion:

* Perceived redundancy
* Many ways to achieve the same result

I suggest we tackle the problem by:

* Getting rid of `filled` and `maybe` methods — those could be replaced with a single `value` with extra predicates and type specs
* Specifying the output type and the constraints in the same place, as opposed to using `type(...).value(...)`
* Renaming `value` into something more apparent, like `output`

solnic:
...so, forget my idea - @flash-gordon reminded me of his idea that I really liked, and this is how it looks now:

``` ruby
required(:age).filter(format?: /\d+/).value(:int, gt?: 18)
```

The type spec is still optional, this is something we have to figure out before the release.

solnic:
@timriley this is now updated and ready for a review

lib/dry/schema/macros/key.rb
```diff
@@ -21,6 +48,15 @@ def to_rule
         def to_ast
           [:predicate, [:key?, [[:name, name], [:input, Undefined]]]]
         end
+
+        private
+
+        def extract_type_spec(*args)
+          type_spec = args[0].is_a?(Symbol) && !args[0].to_s.end_with?(QUESTION_MARK) && args[0]
```
timriley: OK, so I like what we're doing here by allowing the first arg of `value`/`filled`/`maybe` to act as the type, but it seems we have some creeping inconsistencies here?

When we use the _explicit_ `type` API, e.g.

```ruby
required(:age).type(:int).filled(gt?: 18)
```

We allow _any_ value to be passed to `#type` and then it's just directly passed to `schema_dsl.set_type` from there.

So this behaviour on `type` would also allow something like this, I'm assuming:

```ruby
required(:age).type(Types::MyCustomIntegerType).filled(gt?: 18)
```

Am I right with that?

So then we come to this alternative approach to specifying types, via `#extract_type_spec, where we infer a type specification from the first arg passed to `#value`, we have a much stricter set of rules, i.e. it _must_ be a symbol not ending in a question mark.

This would obviously prevent e.g. `MyCustomIntegerType` from being used as the first arg here. And I presume it'd also prevent type specs being passed like `[:nil, MyCustomIntegerType]` where we want a value to be nullable.

Would it make sense here to tweak the logic for this arg check a little, perhaps to something like this?

```ruby
type_spec = args[0]
type_spec = nil if type_spec.is_a?(Symbol) && type_spec.to_s.end_with?(QUESTION_MARK)
```

I'm sure we can find a more succinct way to do this in code, but I think this naive first pass actually expresses the _logic_ pretty clearly, i.e.

1. Let's presume the first arg is going to be a type spec (this is what'll let us permit _any_ kind of object acting as the spec)
2. But if it turns out to be a symbol ending with a question mark, let's take it back, that's not a type spec, that's actually just an argument-less predicate

So this would permit both `value(:int, gt?: 18)` _and_ `value(Types::MyCustomInteger, gt?: 18)`, as well as any other type spec format.

Does... that make sense?

But I also wonder, does permitting the first arg to `#value` to be a type spec, especially if we loosen the arg check like I suggest, actually render the entire `#type` method redundant? 

I guess `#type` is there for (a) when we're not actually providing any `#value` rules for a key, and (b) making things more explicit/clearer by letting us separate the type spec definition rather than mixing it with the predicates (i.e. more a stylistic thing more than anything else)

Am I overlooking anything in the nuances of handling types here?

lib/dry/schema/result.rb
```diff
@@ -12,10 +12,38 @@ class Result
...
 
       option :message_compiler
 
+      def self.new(*args)
+        result = super
+        yield(result)
+        result.freeze
+      end
+
+      def set(hash)
```
timriley: I wonder if `replace` is a more intention-revealing name here?

I think it would certainly help me understand the usage inside `Processor#call` a little better, e.g.

```ruby
result.replace(output) if output.is_a?(::Hash)
```

This would give us a nice naming alignment with the Ruby core language, e.g. `Array#replace` and `Hash#replace`.

lib/dry/schema/macros/key.rb
timriley: FWIW, I read back over our conversation about this in Gitter, and it seems like we came to the conclusion that _this_ was a good DSL:

```ruby
required(:age).filter(format?: /\d+/).value(:integer, gt?: 18)
```

And that it would effectively be a _synonym_ for:

```ruby
required(:age).filter(format?: /\d+/).type(:integer).value(gt?: 18)
```

So I think this is good motivation for us to make our type handling consistent across both variations of that expression, i.e. allow the same type spec arg in both.

lib/dry/schema/key_coercer.rb
```diff
@@ -22,7 +22,7 @@ def initialize(key_map, &coercer)
       end
       
       def call(source)
-        key_map.write(source)
+        key_map.write(Hash(source))
```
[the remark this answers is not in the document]
solnic: @timriley source can be an instance of `Result`, see `Processor#call`

lib/dry/schema/macros/key.rb
solnic: @timriley this was a major oversight on my behalf! thanks for spotting this, it's fixed now

solnic:
> @solnic I love this! Really solid work, it was an inspiring PR to read through.

oh thank you ❤️ 

> Am I right to assume that our plan would be to extend this work (once its merged) and try to support inferring a type spec from the first of the list of predicates supplied for a value? e.g. for key(:age).required(:int?, gt?: 18), we could strip the ? off :int? and see if the resulting string matches any of our built-in types?

Yes, this is the very first I will do after this is merged.

> I know it's not part of this PR, but this is the first time I noticed it... and I really love what you did to support something like array[:int] being used for a type spec. This is way more intention-revealing than the old [:int] structure we had to use in dry-v. Kudos!

Ah thanks, I'm glad you like it.

> I was wondering if we could take this a step further, and support something like nullable[:int] as a kind of type spec? (maybe[:int] would perhaps be even better but I think that could lead to a bit of confusion given the existence of the maybe value-rules macro?)

Yes this would be a great improvement, it's a matter of deciding on the name of the method. People do confuse element order in the array, ie `[:str, :nil]` is not the same as `[:nil, :str]`.

solnic:
@timriley hah funny you mentioned it because it crossed my mind yesterday when I was updating `:int` => `:integer` 😄 yeah we could rename them, esp that we'll be inferring type predicates from type specs, so people won't have to use them explicitly.

PR #13, opened 2019-01-09
Use a processor for applying coercion and schema rules

solnic:
This adds a processor with multiple steps that are executed when a schema is applied. This is now hardcoded to the following 3 steps:

1. Key coercion handled by the new `KeyCoercer` object. This is actually the biggest addition in this PR. `KeyCoercer` uses an instance of `KeyMap` class, which holds detailed information about the keys and have the ability to build up hashes based on these keys definitions.
2. Value coercion handled by a hash schema from dry-types. This works just like before except that **we no longer use it for key coercion**.
3. Applying rules - the very last step is to apply rules defined within the DSL. Works just like before.

### TODO

- [x] extract key coercer functions into singletons
- [x] DRY-up key map / key classes
- [x] add caching for key maps and keys
- [x] memoize key's coerced names


solnic:
@timriley could you take a look at this one?

lib/dry/schema/dsl.rb
```diff
@@ -84,6 +103,30 @@ def set_type(name, spec)
         types[name] = resolve_type(spec).meta(omittable: true)
       end
 
+      def key_map(types = self.types)
+        keys = types.keys.each_with_object([]) { |a, e|
+          e << key_value(a, types[a])
+        }
+        km = KeyMap.new(keys)
+
+        if hash_type === :symbolized
```
aglushkov: Shouln't it be `==`, same at line 124

lib/dry/schema/key_coercer.rb
```diff
@@ -0,0 +1,30 @@
+require 'dry/core/cache'
+
+module Dry
+  module Schema
+    class KeyCoercer
+      extend Dry::Core::Cache
+
+      TO_SYM = -> v { v.to_sym }.freeze
```
timriley:
```suggestion
      TO_SYM = :to_sym.proc.freeze
```

lib/dry/schema/processor.rb
```diff
@@ -0,0 +1,57 @@
...
+        definition.to_ast
+      end
+
+      # required by Dry::Logic::Rule interface
+      def ast(input)
+        definition.to_ast
+      end
+
+      def key_map
+        @__key_map__ ||= steps.detect { |s| s.is_a?(KeyCoercer) }.key_map
```
timriley: Should this constant reference mean we need a `require "dry/schema/key_coercer"` at the top of the file?

lib/dry/schema/key.rb
```diff
@@ -0,0 +1,111 @@
+module Dry
+  module Schema
+    class Key
+      extend Dry::Core::Cache
+
+      DEFAULT_COERCER = -> v { v }
```
timriley:
```suggestion
      DEFAULT_COERCER = :itself.to_proc.freeze
```

lib/dry/schema/dsl.rb
```diff
@@ -84,6 +103,30 @@ def set_type(name, spec)
         types[name] = resolve_type(spec).meta(omittable: true)
       end
 
+      def key_map(types = self.types)
+        keys = types.keys.each_with_object([]) { |a, e|
+          e << key_value(a, types[a])
+        }
```
timriley: Having single-letter block args here certainly impeded my understanding of this as someone coming in relatively fresh.

Could we change them to `|key_name, arr|` instead? Or something similar?

lib/dry/schema/dsl.rb
timriley: If this is indeed a bug, perhaps we need to add a spec that would have picked this up?

solnic:
> What's the relationship of this to #11? I'm presuming that once this is in, we'd rebase #11 and then finish off the work there?

@timriley yes exactly, I had trouble finishing #11 because it needs a better separation of concerns, which this PR improves.

timriley:
@solnic cool! If you want another set of eyes on #11 once this is in, I'm here 💁‍♂️

PR #14, opened 2019-01-25
Infer type check predicates from type specs

solnic:


lib/dry/schema/macros/key.rb
```diff
@@ -6,6 +6,16 @@ module Dry
   module Schema
     module Macros
       class Key < DSL
+        TYPE_TO_PREDICATE = ::Hash.new { |hash, type|
+          primitive = type.maybe? ? type.right.primitive : type.primitive
+
+          if hash.key?(primitive)
+            hash[primitive]
+          else
+            :"#{primitive.name.downcase}?"
```
timriley: I wonder if we want this hash to act somewhat like a cache, and have it save entries inside itself whenever it finds a match?

lib/dry/schema/macros/key.rb
```diff
@@ -52,19 +62,27 @@ def to_ast
...
           end
 
-          predicates = type_spec ? args[1, -1] : args
+          predicates = Array(type_spec ? args[1, -1] : args)
 
-          type(type_spec) if type_spec
+          if type_spec
+            type(nullable && !type_spec.is_a?(::Array) ? [:nil, type_spec] : type_spec)
+            type_predicate = infer_type_predicate
+            predicates << type_predicate unless predicates.include?(type_predicate)
```
timriley: Would we want to put the type_predicate onto the front of the list of predicates, rather than onto the end? If we were writing this by hand, after all, we'd probably be putting the type predicate at the beginning?

PR #19, opened 2019-01-29
Use nil? instead of none?

solnic:


spec/integration/params/predicates/nil_spec.rb
```diff
@@ -124,7 +124,7 @@
       context 'with filled' do
         subject(:schema) do
           Dry::Schema.Params do
-            required(:foo).filled(:none?)
+            required(:foo).filled(:nil)
```
aglushkov: Shouldn't it be `nil?`  with question mark

Issue #24, opened 2017-12-08
.hints throws an error on empty input


tiev:
I encounter an issue when trying out the library to validate Rails controller params.

```ruby
module ParamSchemas
  UserSchema = Dry::Validation.Form do
    required(:user).schema do
      required(:name).filled(:str?)
      required(:age).filled(:int?, gt?: 18)
    end
  end
end

http_params = {}
params = ActionController::Parameters.new(http_params)
result = ParamSchemas::UserSchema.(params.to_unsafe_h)
result.hints

=> TypeError: no implicit conversion of Symbol into Integer
from /Users/viett/.rvm/gems/ruby-2.4.1/gems/dry-validation-0.11.1/lib/dry/validation/message_set.rb:85:in `[]'
```

This only happens when `http_params` is an empty hash.
When `http_params = { user: '' }`, things work fine.

flash-gordon:
I can confirm this
```ruby
UserSchema = Dry::Schema.Form do
  required(:user).schema do
    required(:name).filled(:str?)
    required(:age).filled(:int?, gt?: 18)
  end
end

UserSchema.({}).hints
1: from /Users/gordon/dev/dry-rb/dry-schema/lib/dry/schema/message_set.rb:83:in `block (2 levels) in hints_map'
TypeError (no implicit conversion of Symbol into Integer)
```

nic-lan:
Hi, i have the same issue when `Dry::Schema.Params`

I found an easy fix with overriding the #messages method to rescue from the TypeError,
but i did check with the dry-run spec

```ruby
class Dry::Schema::Result
  def messages(options = Dry::Schema::EMPTY_HASH)
    message_set(options.merge(hints: true)).dump
  rescue TypeError => e
    raise e unless e.message =~ /no implicit conversion of Symbol into Integer/
    errors
  end
end
```

PR #55, opened 2019-02-11
Update result.rb

nic-lan:
Refers to https://github.com/dry-rb/dry-schema/issues/24

You might want to include the proposed changes as a temporary fix


solnic:
Thanks but this will be properly addressed, no need for temporary workarounds.

nic-lan:
yes. of course, you are right

PR #60, opened 2019-02-16
Refactor Messages to use dynamic template objects

### LINKED ISSUE #27: Retrieve validation messages throw ArgumentError when `$DEBUG` is true
Hi, 

This problem is because ruby throws "too many arguments for format string" when `$DEBUG` is true and placeholders and arguments in string templates are not matched. 

I was on ```dry-validation-0.10.5```

Below, I put together a code snippet that would reproduce the behaviour. In this particular case demonstrated by my code snippet, it failed at the beginning of `message_complier.rb#message_text` when the template was `'is missing'` and the tokens were `{ input: null }` .

I ran into this when running hanami in debug, and this would halt it. 

I just wonder if this is an improvement opportunity. 

Many thanks, 

mytram

```
require 'dry-validation'

$DEBUG = true

schema = Dry::Validation.Schema do
  required(:name).filled
end

result = schema.call(
  email: 'jane@doe.org',
  address: { street: 'Street 1', city: 'NYC', zipcode: '1234' }
)

# this will throw ArguementError as $DEBUG is on
puts "error message: #{result.messages.inspect}"

# .rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/dry-validation-0.10.5/lib/dry/validation/message_compiler.rb:163:in `%': too many arguments for format string (ArgumentError)

```

solnic:
This introduces callable templates for messages that are used instead of `String#%`. This way we get faster template evaluation[1] and useful information like which tokens are needed for a template, which are then extracted from message options, so we don't need the nasty workaround for #27 

[1] 👇🏻

```
            String#%   112.000  i/100ms
Dry::Schema::Messages::Template#call
                       171.000  i/100ms
            String#%      1.142k (± 0.9%) i/s -      5.712k in   5.002857s
Dry::Schema::Messages::Template#call
                          1.708k (± 1.9%) i/s -      8.550k in   5.007776s

Comparison:
Dry::Schema::Messages::Template#call:     1708.0 i/s
            String#%:     1141.8 i/s - 1.50x  slower
```

Closes #27

### DIFF
--- a/lib/dry/schema/messages/template.rb
+++ b/lib/dry/schema/messages/template.rb
@@ -1,0 +1,66 @@
+require 'dry/equalizer'
+
+require 'dry/schema/constants'
+
+module Dry
+  module Schema
+    module Messages
+      # Template wraps a string with interpolation tokens and defines evaluator function
+      # dynamically
+      #
+      # @api private
+      class Template
+        include Dry::Equalizer(:text)
+
+        TOKEN_REGEXP = /%{([\w\d]*)}/
+
+        # !@attribute [r] text
+        #   @return [String]
+        attr_reader :text
+
+        # !@attribute [r] tokens
+        #   @return [Hash]
+        attr_reader :tokens
+
+        # !@attribute [r] evaluator
+        #   @return [Proc]
+        attr_reader :evaluator
+
+        # @api private
+        def self.[](input)
+          new(*parse(input))
+        end
+
+        # @api private
+        def self.parse(input)
+          tokens = input.scan(TOKEN_REGEXP).flatten(1).map(&:to_sym)
+          text = input.gsub('%', '#')
+
+          evaluator = <<-RUBY.strip
+            -> (#{tokens.map { |token| "#{token}:" }.join(", ")}) { "#{text}" }
+          RUBY
+
+          [text, tokens, eval(evaluator, binding, __FILE__, __LINE__ - 3)]
+        end
+
+        # @api private
+        def initialize(text, tokens, evaluator)
+          @text = text
+          @tokens = tokens
+          @evaluator = evaluator
+        end
+
+        # @api private
+        def data(input)
+          tokens.each_with_object({}) { |k, h| h[k] = input[k] }
+        end
+
+        # @api private
+        def call(data = EMPTY_HASH)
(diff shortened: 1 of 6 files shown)


lib/dry/schema/messages/template.rb
```diff
@@ -0,0 +1,66 @@
...
+        attr_reader :evaluator
+
+        # @api private
+        def self.[](input)
+          new(*parse(input))
+        end
+
+        # @api private
+        def self.parse(input)
+          tokens = input.scan(TOKEN_REGEXP).flatten(1).map(&:to_sym)
```
flash-gordon: `.flat_map(&:to_sym)` ;)

lib/dry/schema/messages/template.rb
```diff
@@ -0,0 +1,66 @@
...
+          text = input.gsub('%', '#')
+
+          evaluator = <<-RUBY.strip
+            -> (#{tokens.map { |token| "#{token}:" }.join(", ")}) { "#{text}" }
+          RUBY
+
+          [text, tokens, eval(evaluator)]
```
flash-gordon: you should pass `__FILE__` and `__LINE__ - 3` here so that it'll show up in backtraces. It's a lesser issue here but it's better to do it by default

lib/dry/schema/messages/template.rb
solnic: This actually didn't work

flash-gordon: @solnic haha, right, that's because `flat_map` is actually `map_flat` 🤕

solnic: Oops. Did not know this one. Will tweak it later :)

flash-gordon: your version is fine, though you can improve it with regex w/o capturing: `/(?<=%{)\w[\w\d]*(?=})/`, this will allow you to skip `flatten(1)`
```ruby
tokens = input.scan(TOKEN_REGEXP).map(&:to_sym)
```
also note that atm `TOKEN_REGEXP` allows digit-leading and empty tokens, both will break on `eval`

solnic: @flash-gordon OK I will improve this

PR #61, opened 2019-02-18
Fix issues with rule name translations

### LINKED ISSUE #52: Translation for schema keys
Maybe this is a dumb question but I really didn't find it anywhere in the documentation.

How can I translate a Schema field name? Example:

```ruby
UserValidation = Dry::Validation.Schema do
  configure { config.messages = :i18n }

  required(:age).filled
end
```
When I want the full messages i got the message translated correctly, but not the field name.
`age obrigatório(a)`. I didn't find the translation key for 'age', which should be 'idade' of the translated name.

Maybe the yml file should have something like this?
```yaml
pt_BR:
  errors:
    fields:
      age: Idade
```
Sorry about my english. To be clearer I want to know the translation key to be able to translate the Schema fields. If I have `required(:name).filled` I want to setup my i18n yml file to have 'name' translated in the full messages validations.

Thanks for everything and congratulations for the excelent lib! :+1:

### LINKED ISSUE #57: Namespaced rule name translations cannot be resolved
If I set namespace for my schema and use full messages option, this will not get translated:
```
Schema = Dry::Validation.Params do
  configure do
    config.namespace = :my_namespace
  end

  required(:my_rule).filled
end

en:
  rules:
    my_namespace:
       my_rule: Rule translated

Schema.(my_rule: '').messages(full: true) # should output 'Rule translated must be filled'
```
On the other hand, if i didn't set namespace it will work as intended:
```
en:
  rules:
    my_rule: Rule translated
```

I'm not sure if it's a bug or but looks like one.

solnic:
This fixes issues with rule name translations. Previously translations were under `%{locale}.errors.rules`, now they are under root, so `%{locale}.rules`.

Closes #52 
Closes #57

### DIFF
--- a/lib/dry/schema/messages/abstract.rb
+++ b/lib/dry/schema/messages/abstract.rb
@@ -71,10 +71,7 @@ module Dry
         # @api private
         def rule(name, options = {})
           tokens = { name: name, locale: options.fetch(:locale, default_locale) }
-
-          path = rule_lookup_paths.
-            map { |key| key % tokens }.
-            detect { |key| key?(key, options) }
+          path = rule_lookup_paths(tokens).detect { |key| key?(key, options) }
 
           get(path, options) if path
         end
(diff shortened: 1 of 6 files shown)


Mistgun:
Well done dude!

PR #67, opened 2019-02-22
Merging parent rules should not override current rules

### LINKED ISSUE #66: Merging parent rules should not override current rules
https://github.com/dry-rb/dry-schema/blob/master/lib/dry/schema/dsl.rb#L266

it should merge current rules into parent rules, potentially overriding​ some parent rules.

skryukov:
Fixes #66

### DIFF
--- a/lib/dry/schema/dsl.rb
+++ b/lib/dry/schema/dsl.rb
@@ -263,7 +263,7 @@ module Dry
       #
       # @api protected
       def rules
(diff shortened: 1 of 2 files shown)


solnic:
Thanks, but it would be good to add a spec covering this too. Would you be able to do that too?

skryukov:
@solnic no problem, hope this will do =)

PR #68, opened 2019-02-22
Sync with new schema types

flash-gordon:
It took a while but only a few changes are actually required. @solnic dry-schema is way easier to follow/debug compared to dry-validation, that's inspiring 🎉


lib/dry/schema/value_coercer.rb
```diff
@@ -18,7 +18,7 @@ def call(input)
         if input.success?
           type_schema[Hash(input)]
         else
-          type_schema.member_types.reduce(EMPTY_HASH.dup) do |hash, (name, type)|
+          type_schema.name_key_map.reduce(EMPTY_HASH.dup) do |hash, (name, type)|
```
solnic: Maybe it should be an enumerable, so that we could do `type_schema.reduce`?

flash-gordon: yeah, I thought about it, it makes perfect sense now. I'll do this once I finish with testing

PR #80, opened 2019-03-02
Use dry-types from master

flash-gordon:
Changes for dry-types 0.15

### DIFF
--- a/lib/dry/schema/macros/hash.rb
+++ b/lib/dry/schema/macros/hash.rb
@@ -12,7 +12,34 @@ module Dry
         # @api private
         def call(*args, &block)
           trace << hash?
-          super(*args, &block)
+
+          super(*args) unless args.empty?
+
+          if block
+            definition = schema_dsl.new(&block)
+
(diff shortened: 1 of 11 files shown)


lib/dry/schema/macros/hash.rb
```diff
@@ -12,7 +12,34 @@ class Hash < Schema
...
+                schema_type.optional
+              else
+                schema_type
+              end
+
+            schema_dsl.set_type(name, final_type)
+
+            trace << definition.to_rule
+          end
+
```
solnic: This looks like a mistake in conflict resolution, as `Macros::Schema` should include this whole thing whereas `Macros::Hash` simply [prepends `hash?` check and calls super](https://github.com/dry-rb/dry-schema/blob/master/lib/dry/schema/macros/hash.rb#L14-L16).

Specs are passing because the code does the same thing, it's just that it's duplicated now in this branch.

flash-gordon: totally! fixed

flash-gordon:
done via GH

PR #109, opened 2019-03-27
Change interface of Message to allow arbitrary objects

flash-gordon:
This is required by https://github.com/dry-rb/dry-validation/pull/499
Now it's possible to pass custom object from validation rules as errors. Strictly speaking, this wasn't necessary but we don't want those custom objects to be accidentally coerced to strings when you call `MessageSet#to_h` hence this change.


solnic:
Closing in favor of #113

PR #113, opened 2019-03-29
Add support for meta data in Message

### LINKED ISSUE #39: Detailed errors for API's
It would be pretty cool if we could define how errors should be build. I'm talking about something like that: https://stackoverflow.com/a/19673344/3837806

solnic:
This adds support for setting up additional meta data that should be part of a message, ie:

``` yaml
en:
  dry_schema:
    filled?:
      text: "cannot be empty"
      code: 102
```

Closes #39


Mistgun:
<333

PR #124, opened 2019-04-30
Append or skip existing locales

Mistgun:
Append locales only if they do not exist yet to avoid duplicates.


solnic:
Nice catch, TY for the PR. I'm pretty sure it used to be a `Set` so it ensured uniqueness of the values. Oh well 🤷‍♂

Mistgun:
Np. I catched it by accident while testing some locales settings 😆

PR #133, opened 2019-05-12
Fix handling of cache keys for message templates

### LINKED ISSUE #132: `Dry::Schema::Messages::Abstract#call` may return the wrong result in the event of a hash collision.
**Describe the bug**

`Dry::Schema::Messages::Abstract#call`internally caches results based purely on a hash of its arguments. If there is a hash collision then a result may be returned which was calculated for a different set of arguments.

**To Reproduce**

The easiest way to trigger the problem would be to redefine `Hash#hash` to return a more restricted set of values. For example the implementation of `Hash#hash` in Rubinius is roughly
```
def hash
  val = size
  each_pair do |k,v|
    val ^= k.hash
    val ^= v.hash
  end
  val
end
```

**Expected behavior**

A cached value for a different set of arguments should never be returned. If the concern is that the cache might prevent the arguments from being garbage collected then it would be better to cache against `WeakRef.new(args)`.

I will put a change into TruffleRuby to reduce the chance of collisions in our implementation, but that is not a complete fix as collisions could still occur.

**Your environment**

- Ruby version: TruffleRuby RC17
- OS: Linux

solnic:
Fixes #132

### DIFF
--- a/lib/dry/schema/messages/abstract.rb
+++ b/lib/dry/schema/messages/abstract.rb
@@ -45,6 +45,8 @@ module Dry
           String => 'string'
         )
 
+        CACHE_KEYS = %i[path message_type val_type arg_type locale].freeze
+
         # @api private
         def self.cache
           @cache ||= Concurrent::Map.new { |h, k| h[k] = Concurrent::Map.new }
@@ -72,11 +74,6 @@ module Dry
         end
 
         # @api private
-        def hash
-          @hash ||= config.hash
-        end
-
-        # @api private
         def translate(key, locale: default_locale)
           t["#{config.top_namespace}.#{key}", locale: locale]
         end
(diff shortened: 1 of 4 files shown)


lib/dry/schema/messages/abstract.rb
```diff
@@ -95,14 +92,24 @@ def rule(name, options = {})
         # @return [Template]
         #
         # @api public
-        def call(*args)
-          cache.fetch_or_store(args.hash) do
-            text, meta = lookup(*args)
+        def call(predicate, options)
+          cache.fetch_or_store(cache_key(predicate, options).hash) do
```
thedarkone: Hi @solnic, this is still an incorrect fix to the issue raised by @aardvark179 in #132.

The problem is that 2 distinct Ruby objects (in this case arrays returned by the `cache_key` method) are allowed to have identical `hash` values.

The simplest fix for the issue is:
```diff
-cache.fetch_or_store(cache_key(predicate, options).hash) do
+cache.fetch_or_store(cache_key(predicate, options)) do
```

Are you worried that `predicate` and `options` wouldn't get GCed? Why are you using `hash` of the hash-key-object (instead of using the the hash-key-object itself as a key)?

---

```ruby
a = [1,2,3]
b = [2,3,4]

def a.hash; 138293478450784465; end # valid `hash` implementation, this can really happen
def b.hash; 138293478450784465; end # valid `hash` implementation, this can really happen

HASH_RIGHT = {}
HASH_RIGHT[a] = a
HASH_RIGHT[b] = b

HASH_WRONG = {}
HASH_WRONG[a.hash] = a
HASH_WRONG[b.hash] = b

puts HASH_RIGHT.inspect # {[1, 2, 3]=>[1, 2, 3], [2, 3, 4]=>[2, 3, 4]}
puts HASH_WRONG.inspect # {138293478450784465=>[2, 3, 4]}
```
---

Since some time now ruby randomizes its primitives' internal `hash` values by a private `seed` so that each new invocation of ruby VM produces different values for e.g. `puts 1.hash`. What this results in is that there is a __random probability__ for hash collision, any failures will not be reproducible and testing in general will not discover this issue.

solnic: @thedarkone thanks, we've discovered that too and addressing it in #137

PR #141, opened 2019-05-22
Fix filled macro

### LINKED ISSUE #134: filled(:integer) returns "int?" error instead of "filled?"
**Bug**

`filled(:integer)` and `filled(:string)` should return the same error message when input is empty

Actual behaviour:

`filled(:string)` returns "must be filled" (`filled?` error)
`filled(:integer)` returns "must be an integer"  (`int?` error)

**To Reproduce**

```ruby
  required(:user_id).filled(:integer)
end

errors = schema.call(email: '', user_id: '').errors
errors.to_h # => {:email=>["must be filled"], :user_id=>["must be an integer"]}
```

**Expected behavior**

`filled(:integer)` and `filled(:string)` should behave the same and return `filled?` error

**gem version**

```ruby
Dry::Schema::VERSION # => "1.0.2"
```

solnic:
Fixes #134 

This fixes `filled` macro so that it applies `filter(:filled?)` for primitives that are not supported by the corresponding [filled?](https://github.com/dry-rb/dry-logic/blob/master/lib/dry/logic/predicates.rb#L41-L43) predicate. In practice, this means that using `filled` with a type that should be checked if it's filled **before coercion**, will result in applying `filter(:filled?)`.

In example:

``` ruby
dry-schema> s = params { required(:age).filled(:integer) }
=> #<Dry::Schema::Params keys=["age"] rules={:age=>"key?(:age) AND key[age](int?)"}>
dry-schema> s.(age: "").errors.to_h
=> {:age=>["must be filled"]}
dry-schema> s.(age: "foo").errors.to_h
=> {:age=>["must be an integer"]}
dry-schema> s.(age: "30").errors.to_h
=> {}
dry-schema> s.(age: "30").to_h
=> {:age=>30}
```

`JSON` is configured to apply type checks to `nil` instead of `filled?`, so ie:

```ruby
dry-schema> s = json { required(:age).filled(:integer) }
dry-schema> s.(age: nil).errors.to_h
=> {:age=>["must be an integer"]}
```

Contrast that with `Params`:

```
dry-schema> s = params { required(:age).filled(:integer) }
dry-schema> s.(age: nil).errors.to_h
=> {:age=>["must be filled"]}
```

The reason for this different behaviours is that in case of `Params` empty strings are turned into `nil`, which means that a value wasn't filled in, so `filled?` should be applied and provide a useful error message. In case of `JSON` it does not make sense because `nil` means that `null` was passed as the value and it's not allowed because we expect integers. One thing to notice is that using `filled` with numbers in `JSON` schemas is a bit silly, a type check should be enough.

### DIFF
--- a/lib/dry/schema/macros/filled.rb
+++ b/lib/dry/schema/macros/filled.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'dry/schema/primitive_inferrer'
 require 'dry/schema/macros/value'
 
 module Dry
@@ -9,12 +10,22 @@ module Dry
       #
       # @api private
       class Filled < Value
+        # @!attribute [r] primitive_inferrer
+        #   PrimitiveInferrer used to get a list of primitive classes from configured type
+        #   @return [PrimitiveInferrer]
+        #   @api private
+        option :primitive_inferrer, default: proc { PrimitiveInferrer.new }
+
         # @api private
         def call(*predicates, **opts, &block)
           ensure_valid_predicates(predicates)
 
           if opts[:type_spec]
-            value(predicates[0], :filled?, *predicates[1..predicates.size - 1], **opts, &block)
+            if filter?
+              value(*predicates, **opts, &block)
+            else
+              value(predicates[0], :filled?, *predicates[1..predicates.size - 1], **opts, &block)
+            end
           else
             value(:filled?, *predicates, **opts, &block)
           end
@@ -32,6 +43,26 @@ module Dry
           end
         end
         # rubocop:enable Style/GuardClause
+
+        # @api private
+        def filter?
+          !primitives.include?(NilClass) && processor_config.filter_empty_string
+        end
+
+        # @api private
+        def processor_config
+          schema_dsl.processor_type.config
+        end
+
+        # @api private
+        def primitives
+          primitive_inferrer[schema_type]
+        end
+
+        # @api private
+        def schema_type
+          schema_dsl.types[name]
+        end
       end
     end
   end
--- a/lib/dry/schema/primitive_inferrer.rb
+++ b/lib/dry/schema/primitive_inferrer.rb
@@ -1,0 +1,98 @@
+# frozen_string_literal: true
+
+require 'dry/core/cache'
+
+module Dry
+  module Schema
+    # PrimitiveInferrer is used internally by `Macros::Filled`
+    # for inferring a list of possible primitives that a given
+    # type can handle.
+    #
+    # @api private
+    class PrimitiveInferrer
+      extend Dry::Core::Cache
+
+      # Compiler reduces type AST into a list of primitives
+      #
+      # @api private
+      class Compiler
+        # @api private
+        def visit(node)
+          meth, rest = node
+          public_send(:"visit_#{meth}", rest)
+        end
+
+        # @api private
+        def visit_nominal(node)
+          type, _ = node
+          type
+        end
+
+        # @api private
+        def visit_hash(_)
+          Hash
+        end
+
+        # @api private
+        def visit_array(_)
+          Array
+        end
+
+        # @api private
+        def visit_lax(node)
+          visit(node)
+        end
+
+        # @api private
+        def visit_constructor(node)
+          other, * = node
+          visit(other)
+        end
+
+        # @api private
+        def visit_enum(node)
+          other, * = node
+          visit(other)
+        end
+
+        # @api private
+        def visit_sum(node)
+          left, right = node
+
+          [visit(left), visit(right)].flatten(1)
+        end
+
+        # @api private
+        def visit_constrained(node)
+          other, * = node
+          visit(other)
+        end
+
+        # @api private
+        def visit_any(_)
+          Object
+        end
+      end
+
+      # @return [Compiler]
+      # @api private
+      attr_reader :compiler
+
+      # @api private
+      def initialize
+        @compiler = Compiler.new
+      end
+
+      # Infer predicate identifier from the provided type
+      #
+      # @return [Symbol]
+      #
+      # @api private
+      def [](type)
+        self.class.fetch_or_store(type.hash) do
+          Array(compiler.visit(type.to_ast)).freeze
+        end
+      end
+    end
+  end
+end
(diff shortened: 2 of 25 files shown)


lib/dry/schema/macros/filled.rb
```diff
@@ -20,11 +29,36 @@ def call(*predicates, **opts, &block)
           end
 
           if opts[:type_spec]
-            value(predicates[0], :filled?, *predicates[1..predicates.size - 1], **opts, &block)
+            if filter?
+              parent.filter(:filled?)
```
flash-gordon: `parent` could query the child and call `filter` on itself before combining traces

flash-gordon: Something like this (specs passed):
```
diff --git a/lib/dry/schema/macros/dsl.rb b/lib/dry/schema/macros/dsl.rb
index eeed4d0..e3b6da6 100644
--- a/lib/dry/schema/macros/dsl.rb
+++ b/lib/dry/schema/macros/dsl.rb
@@ -66,9 +66,13 @@ module Dry
         # @return [Macros::Core]
         #
         # @api public
-        def filled(*args, &block)
+        def filled(*args, **opts, &block)
           append_macro(Macros::Filled) do |macro|
-            macro.call(*args, &block)
+            defined = macro.call(*args, **opts, &block)
+
+            filter(:filled?) if opts[:type_spec] && defined.filter?
+
+            defined
           end
         end
 
diff --git a/lib/dry/schema/macros/filled.rb b/lib/dry/schema/macros/filled.rb
index 27ecfb0..fdf85c2 100644
--- a/lib/dry/schema/macros/filled.rb
+++ b/lib/dry/schema/macros/filled.rb
@@ -30,7 +30,6 @@ module Dry
 
           if opts[:type_spec]
             if filter?
-              parent.filter(:filled?)
               value(*predicates, **opts, &block)
             else
               value(predicates[0], :filled?, *predicates[1..predicates.size - 1], **opts, &block)
```

lib/dry/schema/primitive_inferrer.rb
```diff
@@ -0,0 +1,98 @@
...
+        def visit_enum(node)
+          other, * = node
+          visit(other)
+        end
+
+        # @api private
+        def visit_sum(node)
+          left, right = node
+
+          [visit(left), visit(right)].flatten
```
flash-gordon: I would write `flatten(1)` because it should suffice here. If it's not working there must be something wrong

lib/dry/schema/primitive_inferrer.rb
```diff
@@ -0,0 +1,98 @@
...
+require 'dry/core/cache'
+
+module Dry
+  module Schema
+    # PrimitiveInferrer is used internally by `Macros::Filled`
+    # for inferring a list of possible primitives that a given
+    # type can handle.
+    #
+    # @api private
+    class PrimitiveInferrer
```
flash-gordon: You know, we could even have it in dry-types, wdyt?

lib/dry/schema/macros/filled.rb
solnic: that's a nice idea, thanks!

lib/dry/schema/primitive_inferrer.rb
solnic: I think it can be deeply nested when there are deeply nested sums like `Integer | String | Array`, no?

flash-gordon: yes but every level has to be balanced with one `visit_sum`, as in `Integer | String | Array` will make two `visit_sum` calls (by the number of `|`)

lib/dry/schema/primitive_inferrer.rb
solnic: Yes, I would like to see this at some point. There are probably more useful tools we could build into dry-types.

lib/dry/schema/macros/filled.rb
```diff
@@ -20,11 +27,35 @@ def call(*predicates, **opts, &block)
...
+            end
           else
             value(:filled?, *predicates, **opts, &block)
           end
         end
+
+        # @api private
+        def filter?
+          !primitives.include?(NilClass) && processor_config.filter_empty_string
+        end
```
solnic: @flash-gordon @timriley what this means is: "if `nil` is NOT an allowed value, and empty strings should be filtered out, then we want to apply `filter(:filled?)`"

lib/dry/schema/processor.rb
```diff
@@ -177,6 +178,20 @@ def rule_applier
         @rule_applier ||= steps.last
       end
       alias_method :to_rule, :rule_applier
+
+      # Check if there are filter rules
+      #
+      # @api private
+      def filter_rules?
+        steps.select { |s| s.is_a?(RuleApplier) }.size > 1
```
flash-gordon: hm, `steps.any?`?

lib/dry/schema/processor.rb
```diff
@@ -177,6 +178,20 @@ def rule_applier
...
+      end
+
+      # Check if there are filter rules
+      #
+      # @api private
+      def filter_schema
+        @filter_schema ||= steps.select { |s| s.is_a?(RuleApplier) }.first
```
flash-gordon: `steps.find`/`steps.detect`?

lib/dry/schema/primitive_inferrer.rb
```diff
@@ -0,0 +1,98 @@
...
+
+        # @api private
+        def visit_nominal(node)
+          type, _ = node
+          [type]
+        end
+
+        # @api private
+        def visit_hash(_)
+          Hash
```
flash-gordon: `visit_*` methods should always return an array, otherwise, `Filled#primitives` may return a class instead of an array of classes

lib/dry/schema/processor.rb
solnic: @flash-gordon I know it's weird but it's actually correct - if there are more than two rule appliers, it means we've got a filter schema and the final schema

lib/dry/schema/macros/filled.rb
timriley: I wonder if we can make this clearer (and less needing explanatory comment like the one you left here) by making some of the names more self-descriptive?

e.g. `def handle_via_filter?` for this method, and `expected_primitives` instead of `primitives`?

PR #150, opened 2019-06-03
Add support for registering custom types

Morozzzko:
So [I wanted this feature](https://dry-rb.zulipchat.com/#narrow/stream/191662-general/topic/dry-validation/near/167066346) and I tried to implement it. 

The implementation is kind of straightforward:

1. Instead of using the global processor's type registry, we're instantiating a new registry for each DSL instance. It damages the total RAM usage, but it enables us to configure types per-instance
2. Since we no longer use global registries, I've renamed `Processor`'s `type_registry` to `type_registry_namespace`.
3. `config` has a new field now: `types`. I was thinking about renaming it to `type_container`, WDYT?
4. I've introduced a new class: `Dry::Schema::TypeContainer`, which is a simple wrapper around `Dry::Types` container. We need it for a couple of reasons:
* `Dry::Types` quacks *kind of* like a container. Except it has `registered?` instead of `key?`. Don't want to change the dependency — it would break backwards compatibility
* I don't want people to manually run `container.merge(Dry::Types.container)` as it exposes internal API, nor do I want them to alias `registered?` to `key?` for compatibility

So the idiomatic code would look something like this:

```ruby
module MyFancyProject
  SchemaTypes = Dry::Schema::TypeContainer.new
  SchemaTypes.register('params.my_string', Types::MyString)
end

Schema = Dry::Validation.Params do
  config.types = MyFancyProject::SchemaTypes
  optional(:foo).filled(:my_string)
end
```

### DIFF
--- a/lib/dry/schema/config.rb
+++ b/lib/dry/schema/config.rb
@@ -5,6 +5,7 @@ require 'dry/configurable'
 
 require 'dry/schema/constants'
 require 'dry/schema/predicate_registry'
+require 'dry/schema/type_container'
 
 module Dry
   module Schema
@@ -26,6 +27,15 @@ module Dry
       # @api public
       setting(:predicates, Schema::PredicateRegistry.new)
 
+      # @!method types
+      #
+      # Return configured container with extra types
+      #
+      # @return [Hash]
+      #
+      # @api public
+      setting(:types, Dry::Types)
+
       # @!method messages
       #
       # Return configuration for message backend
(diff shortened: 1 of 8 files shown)


solnic:
Thanks. I’ll be going through this after dry-validation 1.0.0 is released. First things first.

solnic:
This is really cool, thank you. I'll make it part of `1.2.0`, especially that I've fixed some issues with custom constructor types 😅

PR #163, opened 2019-06-25
Pass type container to nested schemas

flash-gordon:
I managed to do it by passing config. I don't think we need to `dup` it, it'd be strange.


solnic:
This is used in read-only mode so as you said, no need to dup it.

PR #172, opened 2019-07-12
make Monads extension to wrap schema ouput

iarie:
I found out that `Monads` extension is actually not working as it described in [docs](https://dry-rb.org/gems/dry-schema/extensions/monads/)
It _double wraps_ the output which is not very useful:
```ruby
schema = Dry::Schema.Params { required(:name).filled(:str?, size?: 2..4) }
schema.call(name: 'Jane').to_monad
# => Success(#<Dry::Schema::Result{:name=>"Jane"} errors={}>)
schema.call(name: '').to_monad
# => Failure(#<Dry::Schema::Result{:name=>""} errors={:name=>["must be filled"]}>
```

I've changed few lines but I'm concerning of one thing:: should there be `errors.to_h` for `Failure` or just `errors` (instance of `Dry::Schema::MessageSet` class itself)

### DIFF
--- a/lib/dry/schema/extensions/monads.rb
+++ b/lib/dry/schema/extensions/monads.rb
@@ -19,9 +19,9 @@ module Dry
       # @api public
       def to_monad
         if success?
-          Success(self)
+          Success(output)
         else
-          Failure(self)
+          Failure(errors.to_h)
         end
       end
     end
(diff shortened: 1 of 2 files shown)


flash-gordon:
It was a bug in the docs, we just fixed it. The extension always wraps whole `Result` because its API is richer than simple hash output. Making this API not available when you use do notation extensively soon becomes a major pain point. Also, we can't merge this because it's a breaking change.

PR #177, opened 2019-08-03
Reject attempts to build a nested schema for array types built on Dry::Types::Nominal

flash-gordon:
Considering latest changes in dry-types (not released yet) it shouldn't happen very often since the only way to construct `Nominal<::Array>` is to build it manually. Still, it's better to raise a meaningful error than deal with rather unexpected (and inexplicable for many) behavior.


spec/unit/dry/schema/primitive_inferrer_spec.rb
```diff
@@ -11,6 +11,8 @@ def type(*args)
     args.map { |name| Dry::Types[name.to_s] }.reduce(:|)
   end
 
+  let(:types) { Module.new.include(Dry.Types) }
```
solnic: There's a `Types` module defined in spec_helper

PR #187, opened 2019-09-09
Fix array predicate regression

### LINKED ISSUE #186: ArgumentError when upgrading to 1.3.3
**Describe the bug**
I was using 1.3.2 and my code was working, when I upgraded to 1.3.3 I started getting this error:
```
     Failure/Error: required(:service_conforms_to) { array? { each(:str?) } }

     ArgumentError:
       wrong number of arguments (given 0, expected 1)
```

**To Reproduce**

I have a schema like this:

```ruby

ServicesSchema = Dry::Schema.Params do
  required(:service_conforms_to) { array? { each(:str?) } }
end
```

**Expected behavior**

No ArgumentError when upgrading past 1.3.2

**Your environment**

- Affects my production application: no
- Ruby version: 2.6.4
- OS: OsX/Linux

skryukov:
Renamed method to not interfere with the dry-logic predicate. But I can't think of the way to prevent errors like this with tests.

Fixes #186

### DIFF
--- a/lib/dry/schema/macros/schema.rb
+++ b/lib/dry/schema/macros/schema.rb
@@ -28,7 +28,7 @@ module Dry
           definition = schema_dsl.new(&block)
           schema = definition.call
           type_schema =
-            if array?(parent_type)
+            if array_type?(parent_type)
               build_array_type(parent_type, definition.type_schema)
             elsif redefined_schema?(args)
               parent_type.schema(definition.types)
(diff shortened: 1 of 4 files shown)


skryukov:
@solnic thanks for the review!

I added two smoke tests to `integration/params/predicates/array_spec.rb` and `integration/schema/predicates/array_spec.rb`. I think we can do the same for all predicate specs. This will serve as an example for newly added predicates.

solnic:
> I think we can do the same for all predicate specs. This will serve as an example for newly added predicates.

I'll address this separately.

Issue #40, opened 2016-10-18
Explicitly mark that a param can be ignored if nil


TiteiKo:
Given there's a search form, where all fields are optional, and a validation schema that receives the form raw data via rack.

As we know, the form will be submitted with empty strings as values for empty fields.

Extremely simplified, the params could look something like:

``` ruby
Params = Dry::Validation.Form do
  optional(:date).maybe(:date?) # or required, as it's a form, it will be set anyway
end
```

So, when the form submits "" for the date param, the validation output will look like `{ date: nil }`, and we have to handle this nil value somewhere.

Is there any way to explicitly say that this param can be ignored if nil? Like "ok, I know it can be empty or nil, but if it is, please just don't put it in the output", so that the output would look like `{}` and we'd not have to handle the nil value elsewhere.

waiting-for-dev:
@solnic I also need this feature. You said there is some infrastructure prepared. Is is still true for 1.0? Let me know if I can help with this issue.

solnic:
@waiting-for-dev schemas are `Schema::Processor` sub-classes, when you apply a schema to an input what actually happens is that all the steps that a schema has configured will be applied one-by-one in the order they were added. By default [steps are configured like that](https://github.com/dry-rb/dry-schema/blob/master/lib/dry/schema/dsl.rb#L191-L197) based on whatever was defined within the DSL.

So, that's the infrastructure - step-based processors aka schemas. What we're missing is a way of configuring arbitrary steps, so that you could do...let's say this:

```ruby
class MySchema < Dry::Schema::Params
  define do
    required(:name).filled(:string)
  end

  after(:rule_applier) do |input|
    input.compact
  end
end
```

Does this make sense?

solnic:
@waiting-for-dev one more thing - this needs to be cleaned up and promoted to a 1st-class API, right now `Processor` is a bit ugly (you'll see). We need explicit step definitions with unique names (there are built-in ones, like that `rule_applier` in the example above). Then we need a way to access them in a nicer way than doing silly things like `steps.detect { |s| s.is_a?(RuleApplier)`. So overall this requires some refactoring first.

waiting-for-dev:
@solnic I'm afraid I'm a little lost with what is need to be done at first. Which kind of refactor `Processor` exactly  needs? From its docs:

```
    # Processing is split into 4 main steps:
    #
    #   1. Prepare input hash using a key map
[...]

PR #190, opened 2019-09-13
New processor steps interface and before/after step callbacks

skryukov:
This PR implements mentioned in #40 before/after step callbacks and refactors processor steps introducing new class `ProcessorSteps`.

I have some questions:
- should we just ignore callbacks on schema reusing (`required(:foo).schema(FooSchema)`)?
- when a base schema class is used (`Dry::Schema.define(parent: parent)`), should we override callbacks or stack them?
- is the current implementation of `ProcessorSteps` ok? =)

### DIFF
--- a/lib/dry/schema/dsl.rb
+++ b/lib/dry/schema/dsl.rb
@@ -9,6 +9,7 @@ require 'dry/schema/types'
 require 'dry/schema/macros'
 
 require 'dry/schema/processor'
+require 'dry/schema/processor_steps'
 require 'dry/schema/key_map'
 require 'dry/schema/key_coercer'
 require 'dry/schema/value_coercer'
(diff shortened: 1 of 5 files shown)


solnic:
> should we just ignore callbacks on schema reusing (required(:foo).schema(FooSchema))?

We should probably execute them, but I reckon it's gonna be tricky to implement, right?

> when a base schema class is used (Dry::Schema.define(parent: parent)), should we override callbacks or stack them?

Stack them

> is the current implementation of ProcessorSteps ok? =)

Yes 😄

skryukov:
@solnic thanks for your answers!

> We should probably execute them, but I reckon it's gonna be tricky to implement, right?

Yep, I can't even think of the way to implement it for now...

> Stack them

Changed behavior

solnic:
>> We should probably execute them, but I reckon it's gonna be tricky to implement, right?
> Yep, I can't even think of the way to implement it for now...

yeah it's fine to leave it for now

solnic:
@skryukov thanks for working on this. I'm gonna wrap it up and include in 1.4.0.

Issue #42, opened 2018-06-26
Inheriting multiple schemas


artemgurzhii:
I'm having a problem with inheriting multiple schemas at once

Here is an example of what I mean

```rb
Sortable = Dry::Validation.Params do
  optional(:sort).filled
end
```

```rb
Findable = Dry::Validation.Params do
  required(:id).filled(:int?, gt?: 0)
end
```

```rb
Paginatable = Dry::Validation.Params do
  optional(:page).schema do
    required(:number).filled(:int?, gt?: 0)
    required(:size).filled(:int?, gt?: 0)
  end
end
```

```rb
IndexSchema = Dry::Validation.Params(
  ::Schemas::Concerns::Paginatable,
  ::Schemas::Concerns::Sortable,
  ::Schemas::Concerns::Findable
) do
  optional(:filter).schema do
    optional(:email).filled(:str?)
    optional(:fullname).filled(:str?)
  end
end
```

The problem comes from [here](https://github.com/dry-rb/dry-validation/blob/master/lib/dry/validation.rb#L27) as `Params` can have only one base.

waiting-for-dev:
@solnic, about this issue, instead of talking about inheriting multiple schemas, maybe we could just recover custom and inheritable macros, which were available in dry-validation < 1. What do you think? I could try to help with it once I have a bit of time.

PR #191, opened 2019-09-16
Adds support for multiple schema parents

ianwhite:
Re issue #42

This allows an array of parents to be passed as the **parent:** option when creating a schema DSL.

It's a BC change (the schema API for parent: allows an optional DSL (prev API) or array of DSLs).
I've added additional tests for the new behaviour where the parent behaviour was already tested.

The strategy in the solution is to merge all mergeable config with the RHS of the parents overriding the LHS.  In the case of non mergeable config, the last parent is used.

Cheers,
Ian


solnic:
Thank you Ian this is very helpful!

ianwhite:
You're very welcome Piotr.  I'm enjoying dry-rb.

PR #192, opened 2019-09-17
Bug fix: properly memoize @empty in MessageSet

ianwhite:
This tiny pull request fixes that @empty in MessageSet is not properly memoized.


ianwhite:
On reflection, the value is only not memoized when it's cheap to compute, so it's not worth the harder to read code.

PR #202, opened 2019-11-06
Fix all keyword warnings

flash-gordon:


flash-gordon:
I'm merging this. I could have used `*args, **kwargs` for forwading but it does more allocations than `ruby2_keywords`. In the future, `ruby2_keywords` will be replaced with `...` once rubyies < 2.7 are EOL.

PR #217, opened 2019-12-28
Expose schema DSL to `value` and `maybe` with `:hash`

### LINKED ISSUE #215: `value(:hash)` and `maybe(:hash)` should expose schema DSL
`value(:hash) do ... end` and `maybe(:hash) do ... end` should expose nested schema DSL just like `array(:hash)` does. This reduces nesting and is nicer to read.

```ruby
# this
value do
  hash do
    # ...
  end
end

# could be this
value(:hash) do
  # ...
end
```

waiting-for-dev:
Closes #215

### DIFF
--- a/docsite/source/nested-data.html.md
+++ b/docsite/source/nested-data.html.md
@@ -38,20 +38,28 @@ puts errors.to_h.inspect
 # }
 ```
 
+It is equivalent to call `value` macro with `:hash` predicate and a block:
+
+```ruby
+schema = Dry::Schema.Params do
+  required(:address).value(:hash) do
+    # ...
+  end
+end
+```
+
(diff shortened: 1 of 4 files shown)


docsite/source/nested-data.html.md
```diff
@@ -38,20 +38,28 @@ puts errors.to_h.inspect
...
+end
+```
+
 ### Nested Maybe `Hash`
 
 If a nested hash could be `nil`, simply use `maybe` macro with a block:
 
 ```ruby
   required(:address).maybe do
```
flash-gordon: `maybe(:hash)`

waiting-for-dev: Good eyes :+1: Forced a push with it

PR #219, opened 2020-01-06
Add support for pattern matching to Result

flash-gordon:
```ruby
let(:input) { { 'name' => 'John' } }

let(:schema) do
  Dry::Schema.Params { required(:name).filled(min_size?: 3) }
end

example do
  case result
  in { name: } => captured if name.eql?('John')
    name # => "John"
  in { name: other_name }
    # using other_name
  else
    # name not available
  end
end
```

With monads:
```ruby
require 'dry/schema'
require 'dry/monads'

Dry::Schema.load_extensions(:monads)

class Example
  include Dry::Monads[:result]

  Schema = Dry::Schema.Params { required(:name).filled(:string, size?: 2..4) }

  def call(input)
    case schema.(input).to_monad
    in Success(name:)
      "Hello #{name}" # name is captured from result
    in Failure(name:)
      "#{name} is not a valid name"
    end
  end
end

run = Example.new

run.('name' => 'Jane')   # => "Hello Jane"
run.('name' => 'Albert') # => "Albert is not a valid name"
```

### DIFF
--- a/docsite/source/extensions/monads.html.md
+++ b/docsite/source/extensions/monads.html.md
@@ -4,7 +4,7 @@ layout: gem-single
 name: dry-schema
 ---
 
-The monads extension makes `Dry::Schema::Result` objects compatible with `dry-monads`.
+The monads extension makes `Dry::Schema::Result` objects compatible with dry-monads.
 
 To enable the extension:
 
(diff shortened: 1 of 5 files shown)


docsite/source/extensions/monads.html.md
```diff
@@ -29,4 +29,35 @@ schema.(name: "")
...
+```ruby
+require 'dry/schema'
+require 'dry/monads'
+
+Dry::Schema.load_extensions(:monads)
+
+class Example
+  include Dry::Monads[:result]
+
+  Schema = Dry::Schema.Params { required(:name).filled(:str?, size?: 2..4) }
```
solnic: `:str?` => `:string`

PR #216, opened 2019-12-23
Introduce AST error compiler

waiting-for-dev:
It makes message compiler a specific case of a generic error compiler,
just like new ast error compiler.

References dry-rb/dry-validation#604

This is a WIP in regard of adding the possibility to return something different than human readable messages as result errors. It'd be particularly useful in a RESTful scenario when you want your frontend to take care of everything UI-related. I don't want to go any further without having some feedback, as there are some considerations to take into account.

As you can see, I've generalized what it was a `MessageCompiler` in favor of an `ErrorCompiler`, being the former a subclass of the latest. I've added an `AstErrorCompiler` which just returns the plain error ast. Probably it's too raw and we should digest it a little bit, but at the same time it keeps all the flexibility for the machine consuming it.

I've also generalized `MessageSet` and new `AstErrorSet` as two subclasses of `ErrorSet`. As in the `ErrorCompiler` it makes explicit the interface that dry-schema internals expect for them.

As it'd do the error type polymorphic, probably some result methods should be renamed or aliased, as in the case of `#messages`.

PR #221, opened 2020-01-08
Introduce AST error compiler

waiting-for-dev:
From #216 :

It makes message compiler a specific case of a generic error compiler,
just like new ast error compiler.

References dry-rb/dry-validation#604

This is a WIP in regard of adding the possibility to return something different than human readable messages as result errors. It'd be particularly useful in a RESTful scenario when you want your frontend to take care of everything UI-related. I don't want to go any further without having some feedback, as there are some considerations to take into account.

As you can see, I've generalized what it was a `MessageCompiler` in favor of an `ErrorCompiler`, being the former a subclass of the latest. I've added an `AstErrorCompiler` which just returns the plain error ast. Probably it's too raw and we should digest it a little bit, but at the same time it keeps all the flexibility for the machine consuming it.

I've also generalized `MessageSet` and new `AstErrorSet` as two subclasses of `ErrorSet`. As in the `ErrorCompiler` it makes explicit the interface that dry-schema internals expect for them.

As it'd do the error type polymorphic, probably some result methods should be renamed or aliased, as in the case of `#messages`.

### DIFF
--- a/lib/dry/schema/ast_error_compiler.rb
+++ b/lib/dry/schema/ast_error_compiler.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+require 'dry/schema/error_compiler'
+require 'dry/schema/ast_error_set'
+
+module Dry
+  module Schema
+    # Compiles rule results AST into machine-readable format
+    #
+    # @api private
+    class AstErrorCompiler < ErrorCompiler
+      # @api private
+      def call(ast)
+        AstErrorSet[ast]
+      end
+    end
+  end
+end
--- a/lib/dry/schema/ast_error_set.rb
+++ b/lib/dry/schema/ast_error_set.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'dry/schema/error_set'
+
+module Dry
+  module Schema
+    # A set of AST errors used to generate machine-readable errors
+    #
+    # @see Result#message_set
+    #
+    # @api public
+    class AstErrorSet < ErrorSet
+      private
+
+      # @api private
+      def errors_map(errors = self.errors)
+        errors
+      end
+    end
+  end
+end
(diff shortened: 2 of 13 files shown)


solnic:
@waiting-for-dev I tweaked rubocop settings, please rebase

waiting-for-dev:
done @solnic

solnic:
@waiting-for-dev seems like this branch is still far behind master...?

waiting-for-dev:
Damn it!! I started working by error on an old version of master and I fixed it on local in another branch... which is the one I pushed here by mistake. Ok, my bad, now it is the correct branch....

solnic:
@waiting-for-dev why do you think such refactor is needed? If we want a differen representation of the results, it's gonna be much simpler to just add a subclass of `MessageCompiler`.

waiting-for-dev:
I felt that it was the most difficult thing to do but that it would result in a simpler and more decoupled design. I mean, if we just want to return some kind of error identifier we don't need to be concerned and do any kind of work with human readable messages. But maybe I'm missing something about current implementation.

solnic:
> if we just want to return some kind of error identifier we don't need to be concerned nor do any kind of work with human readable messages

@waiting-for-dev OK but these error identifiers must come from *somewhere*. Isn't this just about using an error identifier as the message `text` value, instead of a human-readable text?

solnic:
I'm gonna close this for the time being. Let's revisit it when the time is right (most likely when we start working on 2.0.0). I'm gonna leave the branch on GH for future reference.

waiting-for-dev:
Yeah, sorry @solnic ... I didn't find the time to think more about it... let's revisit it later as you said...

solnic:
@waiting-for-dev no worries man. In general it's better to discuss and plan such refactorings because there's always risk we'll end up with unfinished PRs being opened for too long 🙂

PR #223, opened 2020-01-12
Add each keyword to define arrays as root nodes

### LINKED ISSUE #22: Provide a way of defining schemas for arrays as root nodes
As an example
```ruby
Dry::Schema.JSON do
  each do
    schema do
      required(:name).filled(:str?)
      required(:age).filled(:int?)
    end
  end
end
```

skryukov:
This is proof of concept PR. It adds `each` keyword (`array` is already in use, so I decided to leave it as-is for now), which adds `__root__` key to the schema and input and removes it from the message paths later.

I still have some questions:
1. What message should be returned, if the input is not an array: `MessageSet` is a hash-like value, so what is the key for a "root" message like `input must be an array`?
2. How should we set rules for this kind of schemas in dry-v?
3. How can I set rule for the root array, like `value(:array, min_size?: 1)`?

Closes #22

### DIFF
--- a/lib/dry/schema/dsl.rb
+++ b/lib/dry/schema/dsl.rb
@@ -160,6 +160,32 @@ module Dry
         key(name, macro: Macros::Optional, &block)
       end
 
+      # Define an array
+      #
+      # @see DSL#required
+      #
+      # @example
+      #   each(:integer, gt?: 0)
+      #
+      #   each do
+      #    schema do
+      #      required(:method).filled(:str?)
+      #      required(:amount).filled(:float?)
+      #    end
+      #  end
+      #
+      # @return [Macros::Required]
+      #
+      # @api public
+      def each(*args, **opts, &block)
+        @__root__ = true
+        required(:__root__).array(*args, **opts, &block)
+      end
+
+      def with_root?
+        instance_variable_defined?('@__root__')
+      end
+
       # A generic method for defining keys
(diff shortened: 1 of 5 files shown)


solnic:
How about introducing `Dry::Schema.Array` DSL type instead of trying to extend schema DSL? This feature would become so much simpler to implement, ie we could have:

```ruby
array_schema = Dry::Schema.Array do
  schema do
  end
end
```

it would expose the schema DSL and build a schema from it and then provide custom `#call` that simply goes through array elements and applies the schema and returns indexed results. With such implementation we could easily add additional features like rules for the array itself, ie:

```ruby
array_schema = Dry::Schema.Array do
  input do
    size?(10)
  end

  schema do
  end
end
```

mrbongiolo:
I like the `Dry::Schema::Array` dsl @solnic. Our project has some API endpoints that receive data in arrays, think like a list of prices for a stock. That's why we're still using the old `0.10.0` version, this feature would allow us to upgrade to the latest features.

solnic:
@skryukov would you like to give it a shot?

skryukov:
> @skryukov would you like to give it a shot?

Sure, I'll try my best =)

solnic:
Please open a new PR with `Dry::Schema.Array do ... end`. I'm going to close this one for now. Thank you.

PR #232, opened 2020-02-10
Fix example suggesting conversion to empty array

janko:
I noticed that dry-schema doesn't coerce an empty string to an empty array in this example, it converts it to nil:

```rb
require "dry/schema"
require "dry/schema/version"

Dry::Schema::VERSION # => "1.4.3"

  required(:age).maybe(:integer, gt?: 18)
  required(:tags).maybe(:array)
end

result = schema.call('email' => 'jane@doe.org', 'age' => '', 'tags' => '')

puts result.to_h
# >> {:email=>"jane@doe.org", :age=>nil, :tags=>nil}
```

So I'm just sending a PR correcting this 😃

### DIFF
--- a/docsite/source/params.html.md
+++ b/docsite/source/params.html.md
@@ -31,17 +31,16 @@ puts errors.to_h.inspect
 
 ### Handling empty strings
 
-Your schema will automatically coerce empty strings to `nil` or an empty array, provided that you allow a value to be nil:
+Your schema will automatically coerce empty strings to `nil`, provided that you allow a value to be nil:
 
 ```ruby
 schema = Dry::Schema.Params do
   required(:email).filled(:string)
   required(:age).maybe(:integer, gt?: 18)
-  required(:tags).maybe(:array)
 end
 
-result = schema.call('email' => 'jane@doe.org', 'age' => '', 'tags' => '')
+result = schema.call('email' => 'jane@doe.org', 'age' => '')
 
 puts result.to_h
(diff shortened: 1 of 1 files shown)


solnic:
Actually, it does, but not when `maybe` is used - that's the actual error in the docs, see:

```ruby
dry-schema> schema = Dry::Schema.Params do
dry-schema*>   required(:tags).value(:array)   
dry-schema*> end   
=> #<Dry::Schema::Params keys=["tags"] rules={:tags=>"key?(:tags) AND key[tags](array?)"}>
dry-schema> schema.(tags: "")
=> #<Dry::Schema::Result{:tags=>[]} errors={}>
```

janko:
That's good to know, thanks!

How should I update the docs? Should I just replace the `maybe` in this example? It really makes sense to me now that `maybe` will leave `nil` and `value` will coerce, I could also update the `maybe`/`value` docs to mention this, as I wish I've known that 😃

solnic:
> How should I update the docs? Should I just replace the maybe in this example?

Yes please :)

> It really makes sense to me now that maybe will leave nil and value will coerce, I could also update the maybe/value docs to mention this, as I wish I've known that 😃

That would be lovely!

solnic:
hey @janko we're wrapping up `1.5.0` release and I'm wondering if you're still interested in improving this part of the documentation?

janko:
Hi, at the moment I don't have time to contribute, so feel free to take over.

PR #233, opened 2020-02-14
Delegate caching and parameter interpolation to message backends

### LINKED ISSUE #211: Delegate message parameter interpolation to message backends
Currently, parameter interpolation is handled by `dry-schema` in [Dry::Schema::Messages::Template](https://github.com/dry-rb/dry-schema/blob/master/lib/dry/schema/messages/template.rb). By delegating this responsibility to the message backends themselves, `dry-schema` could take advantage of behaviors specific to a given backend.

## Examples

`I18n` supports loading translations from Ruby files. These Ruby files can contain lambdas as translation values that will be invoked by `I18n`:

```ruby
# en.rb
{
  en: {
    foo: ->(key, options) do
      <<~OUTPUT
        key: #{key.inspect}
        options: #{options.inspect}
      OUTPUT
    end
  }
}

# example.rb
require "i18n"

I18n.load_path << "en.rb"

puts I18n.t(:foo, bar: "baz")
```

```
$ ruby example.rb
key: :foo
options: {:bar=>"baz"}
```

As it stands, `dry-schema` does not pass along parameters when generating a message, due to interpolation being handled outside of the `I18n` backend:

```ruby
# en.rb
{
  en: {
    dry_schema: {
      errors: {
        lt?: lambda do |key, options|
          puts <<~OUTPUT
            key: #{key.inspect}
            options: #{options.inspect}
          OUTPUT

          "must be less than %{num}"
        end
      }
    }
  }
}

# example.rb
require "i18n"
require "dry-schema"
require "pp"

I18n.load_path = ["en.rb"]

  config.messages.backend = :i18n
  config.messages.load_paths = []

  required(:count).value(lt?: 10)
end

pp schema.call(count: 10).errors.to_h
```

```
$ ruby example.rb
key: "dry_schema.errors.lt?"
options: {}
{:count=>["must be less than 10"]}
```

Since the translation is evaluated before it is converted to a message template, `I18n` doesn't receive any the options that will eventually be used to perform the interpolation. If, however, the template were to defer to the backend to perform interpolation, then the options could be passed through at template expansion time.

I'm imagining a situation in which an application calls out to an external CMS or into a specific library to translate messages and it leverages `I18n`'s lambda support to implement this.

I'd be happy to look into implementing this. Thanks for a creating such a great library!

robhanlon22:
Resolves #211


lib/dry/schema/message_compiler.rb
```diff
@@ -177,13 +180,16 @@ def lookup_options(arg_vals:, input:)
       end
 
       # @api private
-      def message_text(template, tokens, options)
-        text = template[template.data(tokens)]
+      def message_text(predicate, tokens, options)
+        result = messages.lookup(predicate, tokens, options)
 
-        return text unless full
+        return result if !result || !full
```
robhanlon22: N.B. this is a bit of a behavior change that I think makes sense—if there is no result, then return that lack of result. Previously, if there was no text, then a string with the rule name followed by a space would still be returned, which would be puzzling

robhanlon22:
@solnic resolved

spec/integration/messages/i18n_spec.rb
```diff
@@ -12,21 +11,24 @@
...
   before do
     I18n.config.available_locales = [:en, :pl]
-    I18n.load_path.concat(%w(en pl).map { |l| SPEC_ROOT.join("fixtures/locales/#{l}.yml") })
+    I18n.load_path.concat(%w[en pl].map { |l| SPEC_ROOT.join("fixtures/locales/#{l}.yml") })
     I18n.backend.load_translations
     I18n.reload!
   end
 
-  describe '#[]' do
+  describe '#lookup' do
```
solnic: @robhanlon22 why did you have to change the spec?

robhanlon22: `#[]`/`#call` returned instances of `Template`, but `Template` was removed in this PR, so `#[]`/`#call` was removed, thus the change to the spec

solnic:
@robhanlon22 another conflict after merging your other PR 😄

robhanlon22:
@solnic resolved again :)

solnic:
@robhanlon22 thanks Rob, this should be a great improvement. I gave you repo access, so if you feel like sending another awesome PR, please use this repo instead of your fork ❤️

PR #236, opened 2020-02-16
Add a few more tests to verify new interpolation behavior

robhanlon22:


spec/integration/messages/i18n_spec.rb
```diff
@@ -181,20 +202,29 @@
         )
       end
 
+      it 'can use a proc for a message' do
+        store_errors(
+          predicate_proc: ->(_path, opts) { opts[:text] }
```
robhanlon22: N.B. I tried to use keyword arguments for `opts` here but it caused warnings on Ruby 2.7.0 :(

PR #237, opened 2020-02-16
Infer schemas from hash types

### LINKED ISSUE #218: Arrays of Hashes breaks in v1.4.2
**Describe the bug**

No code changes, just a dry-schema gem upgrade. Hash schemas have notoriously been problematic with dry-rb libraries and my workarounds break with every few releases. In dry-schema v1.4.1 everything works fine, but v1.4.2 introduces the notorious `NoMethodError: undefined method 'visit_schema' for #<Dry::Types::PredicateInferrer::Compiler:0x00007fdb4eb2c3f8>` error.

**To Reproduce**

This example reproduces the issue where it works in v1.4.1 and breaks in v1.4.2.

```ruby
# Types
T=Class.new{include Dry.Types(default: :nominal)}
# Custom type
UUID_V4=T::Strict::String.constrained(format: /\A[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-4[0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}\z/i).constructor(&:to_s).meta(type: :UUID_V4)
# Hash schema in array
HASHES=T::Strict::Array.of(T::Hash.schema(field: UUID_V4, value: T::Strict::String))
# Contract which throws an error
C=Class.new(Dry::Validation::Contract) { params { required(:hashes).maybe(HASHES) } }
```

There are other cases in which I get this error, but this is one I was able to create a reproducible script for.

Semi-related: I've seen the dry-rb team say in threads elsewhere that complex Hashes aren't very supported.  I've gotten around those kinds of issues by wrapping complex Hash schemas inside a Hash constructor. I'm not doing that in the example above to reproduce the regression, but here's an example how I wrap Hash schemas elsewhere. Have yet to see if this is now also broken.

```ruby
T::Hash.constructor(T::Hash.schema(field: UUID_V4, value: T::String.constructor(&:to_s)))
```

**Expected behavior**

The expected behavior is that the contract doesn't throw an error building the predicates and can be used to validate a hash.

**Your environment**

- Ruby version: 2.5.6p201
- OS: macOS Mojave

flash-gordon:
Closes #218

### DIFF
--- a/lib/dry/schema/macros/value.rb
+++ b/lib/dry/schema/macros/value.rb
@@ -38,6 +38,12 @@ module Dry
 
             if block && type_spec.equal?(:hash)
               hash(&block)
+            elsif type_spec.respond_to?(:keys)
+              hash do
+                type_spec.each do |key|
+                  required(key.name).value(key.type)
+                end
+              end
             elsif block
               trace.append(new(chain: false).instance_exec(&block))
             end
--- a/lib/dry/schema/macros/value.rb
+++ b/lib/dry/schema/macros/value.rb
@@ -41,7 +41,11 @@ module Dry
             elsif type_spec.respond_to?(:keys)
               hash do
                 type_spec.each do |key|
-                  required(key.name).value(key.type)
+                  if key.required?
+                    required(key.name).value(key.type)
+                  else
+                    optional(key.name).value(key.type)
+                  end
                 end
               end
             elsif block
--- a/lib/dry/schema/macros/value.rb
+++ b/lib/dry/schema/macros/value.rb
@@ -23,7 +23,7 @@ module Dry
                 schema.type_schema
               end
 
-            schema_dsl.set_type(name, updated_type)
+            type(updated_type)
           end
 
           trace_opts = opts.reject { |key, _| key == :type_spec || key == :type_rule }
--- a/lib/dry/schema/macros/value.rb
+++ b/lib/dry/schema/macros/value.rb
@@ -39,15 +39,7 @@ module Dry
             if block && type_spec.equal?(:hash)
               hash(&block)
             elsif type_spec.respond_to?(:keys)
-              hash do
-                type_spec.each do |key|
-                  if key.required?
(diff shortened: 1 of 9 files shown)


spec/integration/schema/macros/value_spec.rb
```diff
@@ -217,5 +217,24 @@
...
+          required(:data).value(env.type)
+        end
+      end
+
+      it 'unpacks type' do
+        expect(schema.(data: { name: nil }).messages).to eql(
+          data: { name: ['must be a string'] }
+        )
+      end
+    end
```
solnic: Could you add an example with a deeply nested hash?

flash-gordon: @solnic I'll be adding all sorts of examples, stay tuned

solnic:
@flash-gordon this will address #218 right?

flash-gordon:
@solnic I'm not sure yet, mb arrays will require some additional work

solnic:
@flash-gordon let's get my fix merged first #243 - I think it's gonna be helpful here too

flash-gordon:
@solnic sure 👍

lib/dry/schema/macros/value.rb
```diff
@@ -38,6 +38,16 @@ def call(*predicates, **opts, &block)
...
+            elsif type_spec.respond_to?(:keys)
+              hash do
+                type_spec.each do |key|
+                  if key.required?
+                    required(key.name).value(key.type)
+                  else
+                    optional(key.name).value(key.type)
+                  end
+                end
+              end
```
solnic: @flash-gordon we could tweak `hash` to accept a hash schema so that this becomes `hash(type_spec)`

flash-gordon: sure, thanks for the hint. I actually have no plans leave it as such, just making it work first

flash-gordon:
@solnic I think it's good to go at this point

PR #238, opened 2020-02-16
Remove predicate inferrer

flash-gordon:
It was moved to dry-types and for some reason I didn't remove it before.


solnic:
@flash-gordon could you mention this removal in the changelog? You never know if somebody uses it 🤷‍♂

flash-gordon:
@solnic 🤔 mb we deliberately shouldn't mention it? It was part of private API, and it would be better if the user revealed themself somehow :) I'll add an entry anyway

solnic:
@flash-gordon oh right, I forgot about it. Do what you think is best 😄

PR #239, opened 2020-02-16
Expose global configuration

### LINKED ISSUE #205: Global I18n load_path config
Ability to set a global load_path config on messages so that each instance of the Schema does not require explicit i18n configuration.

## Examples

```
# config/initializers/dry_schema.rb

Dry::Schema.config.messages.load_paths += I18n.load_path
```

Willing to help work this if it's something that dry-schema would be interested in accepting!

robhanlon22:
Resolves #205.

### DIFF
--- a/lib/dry/schema.rb
+++ b/lib/dry/schema.rb
@@ -14,6 +14,18 @@ module Dry
   module Schema
     extend Dry::Core::Extensions
 
+    # Configuration
+    #
+    # @example
+    #   Dry::Schema.config.messages.backend = :i18n
+    #
+    # @return [Config]
+    #
+    # @api public
+    def self.config
+      @config ||= Config.new
+    end
+
     # Define a schema
     #
     # @example
--- a/lib/dry/schema.rb
+++ b/lib/dry/schema.rb
@@ -2,6 +2,7 @@
 
 require 'dry/core/extensions'
 
+require 'dry/schema/config'
 require 'dry/schema/constants'
 require 'dry/schema/dsl'
 require 'dry/schema/params'
--- a/lib/dry/schema/dsl.rb
+++ b/lib/dry/schema/dsl.rb
@@ -2,6 +2,7 @@
 
 require 'dry/initializer'
 
+require 'dry/schema'
 require 'dry/schema/constants'
 require 'dry/schema/config'
 require 'dry/schema/compiler'
--- a/lib/dry/schema/dsl.rb
+++ b/lib/dry/schema/dsl.rb
@@ -67,7 +67,7 @@ module Dry
       option :parent, Types::Coercible::Array, default: -> { EMPTY_ARRAY.dup }, as: :parents
 
       # @return [Config] Configuration object exposed via `#configure` method
-      option :config, optional: true, default: proc { parent ? parent.config.dup : Config.new }
+      option :config, optional: true, default: proc { (parent || Schema).config.dup }
 
       # @return [ProcessorSteps] Steps for the processor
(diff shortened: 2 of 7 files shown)


lib/dry/schema.rb
```diff
@@ -14,6 +14,18 @@ module Dry
...
+    #
+    # @example
+    #   Dry::Schema.config.messages.backend = :i18n
+    #
+    # @return [Config]
+    #
+    # @api public
+    def self.config
+      @config ||= Config.new
+    end
```
solnic: We follow the 'require what you require` rule, which means this file needs `require "dry/schema/config"` at the top.

lib/dry/schema/dsl.rb
```diff
@@ -67,7 +67,7 @@ class DSL
       option :parent, Types::Coercible::Array, default: -> { EMPTY_ARRAY.dup }, as: :parents
 
       # @return [Config] Configuration object exposed via `#configure` method
-      option :config, optional: true, default: proc { parent ? parent.config.dup : Config.new }
+      option :config, optional: true, default: proc { (parent || Schema).config.dup }
```
solnic: Would it be possible to encapsulate this in the parent, so that this line could remain unchanged?

spec/integration/custom_error_messages_spec.rb
```diff
@@ -46,5 +46,21 @@
 
       include_context 'schema with customized messages'
     end
+
+    context 'with global configuration' do
+      before do
+        config = Dry::Schema::Config.new
+        config.messages.backend = :i18n
+        allow(Dry::Schema).to receive(:config).and_return(config)
+      end
```
solnic: We have [a test mode](https://dry-rb.org/gems/dry-configurable/0.11/testing/) for configurables, it would be cleaner and more future-proof if we enabled it.

spec/unit/dry/schema/schema_spec.rb
```diff
@@ -36,4 +36,10 @@
       end
     end
   end
+
+  describe '.config' do
+    it 'returns an instance of Config' do
+      expect(Dry::Schema.config).to be_instance_of(Dry::Schema::Config)
+    end
+  end
```
solnic: This seems redundant to be honest

lib/dry/schema/dsl.rb
robhanlon22: I don't think it's possible, since this line checks for the presence of the parent to determine whether to use the parent's config or the global config.

spec/unit/dry/schema/dsl_spec.rb
```diff
@@ -34,4 +34,32 @@
...
+  describe '#config' do
+    let(:namespace) { :test_namespace }
+    let(:replacement_namespace) { :replacement_test_namespace }
+
+    it "uses a dup of the parent's config when a parent is present" do
+      config = Dry::Schema::Config.new.tap { |c| c.messages.namespace = namespace }
+      parent = Dry::Schema::DSL.new(config: config)
+      dsl = Dry::Schema::DSL.new(parent: parent)
+
+      aggregate_failures do
```
flash-gordon: all failures are aggregated automatically, there's a way to do it
https://github.com/dry-rb/dry-schema/blob/ab17647d5fe62b22ff1edeccf9a934c4eadc9e7d/spec/spec_helper.rb#L64-L66

robhanlon22: Thanks, you learn something new every day :)

PR #197, opened 2019-10-09
Add multiple parents callback support

skryukov:
Adds support for multiple parents callbacks

PR #246, opened 2020-02-20
Add multiple parents callback support

solnic:
Refs #197

### DIFF
--- a/lib/dry/schema/dsl.rb
+++ b/lib/dry/schema/dsl.rb
@@ -71,7 +71,7 @@ module Dry
       option :config, optional: true, default: proc { (parent || Schema).config.dup }
 
       # @return [ProcessorSteps] Steps for the processor
-      option :steps, default: proc { parent ? parent.steps.dup : ProcessorSteps.new }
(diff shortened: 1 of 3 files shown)


robhanlon22:
@solnic There are a couple more references to `parents.last` as the single parent of a particular DSL. Do you think that those should also be made multiple parent-aware to prevent any future surprises?

robhanlon22:
My mistake, there's only once reference left—within the `config` setting that I recently touched

solnic:
@robhanlon22 tbh I'm not sure if we want to inherit config when there's more than one parent. I think people use multiple parents to compose a schema rather than to inherit configuration 🤔

robhanlon22:
I'm actually with you on that—it might be surprising to have your config come from a parent when you're expecting the composition to be more or less stateless.

solnic:
@robhanlon22 another option is to consider checking if all configs are equal and raise if they are not

robhanlon22:
@solnic I feel like that would be the most correct but also the most disruptive. I think removing config inheritance and making it explicit would be best

PR #249, opened 2020-02-22
Restore Messages::Abstract#call

solnic:
This was removed in #233 but it's a public interface that dry-validation
relies on.

Its behavior is now different since it no longer returns a `Template` as
evaluation is handled by the specific backend now (which is a good thing).

This means **we will have to bump version to 2.0.0** already if we decide to release it along with other changes.

### DIFF
--- a/lib/dry/schema/messages/abstract.rb
+++ b/lib/dry/schema/messages/abstract.rb
@@ -80,6 +80,20 @@ module Dry
           rule.is_a?(Hash) ? rule[:text] : rule
         end
 
+        # Retrieve a message template
+        #
+        # @return [Hash]
+        #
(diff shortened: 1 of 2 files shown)


robhanlon22:
I think that we could restore the Template class now if we'd like to avoid bumping to 2.0.0. It would be a pretty thin shim, since it would just proxy to its backend.

solnic:
@robhanlon22 yeah I've thought about it and it would be great to restore the original behavior of `Messages#call`. Bumping to 2.0.0 because of this would mean that we'll have 3.0.0 quickly after because I planned making type specs mandatory *soon* and that's a public API change too.

PR #251, opened 2020-02-27
Reintroduce Template class that delegates to message backends

robhanlon22:


lib/dry/schema/messages/abstract.rb
```diff
@@ -107,21 +122,6 @@ def looked_up_paths(predicate, options)
           filled_lookup_paths(tokens)
         end
 
-        # Try to find a message for the given predicate and its options
-        #
-        # @api private
-        #
-        # rubocop:disable Metrics/AbcSize
-        def lookup(predicate, tokens, options)
```
robhanlon22: I removed this because it was marked as `@private` and it seemed simpler to combine this into `[]`.

lib/dry/schema/messages/namespaced.rb
```diff
@@ -69,6 +69,16 @@ def rule_lookup_paths(tokens)
         def cache_key(predicate, options)
           messages.cache_key(predicate, options)
         end
+
+        # @api private
+        def pruned_data(template, **input)
```
robhanlon22: I wanted to call this `data` but it collides with the existing `data` attr

lib/dry/schema/messages/yaml.rb
```diff
@@ -151,15 +144,24 @@ def prepare
         self
       end
 
-      private
+      # @api private
+      def pruned_data(template, **input)
```
robhanlon22: I wanted to call this `data` but it collides with the existing `data` attr

PR #253, opened 2020-02-29
Add struct extension

flash-gordon:
It simply infers a hash schema from a struct. More complex scenarios are not supported, there are reasons for this. In future, we plan to infer structs from schema definition, similarly to how it works in rom.

### DIFF
--- a/lib/dry/schema/extensions/struct.rb
+++ b/lib/dry/schema/extensions/struct.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+require 'dry/struct'
+require 'dry/schema/predicate_inferrer'
+require 'dry/schema/primitive_inferrer'
+require 'dry/schema/macros/dsl'
+require 'dry/schema/macros/hash'
+
+module Dry
+  module Schema
+    module Macros
+      Hash.prepend(::Module.new {
+        def call(*args)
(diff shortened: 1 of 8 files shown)


solnic:
> More complex scenarios are not supported

@flash-gordon could you give a couple of examples? 🤔

lib/dry/schema/extensions/struct.rb
```diff
@@ -0,0 +1,31 @@
...
+require 'dry/schema/macros/hash'
+
+module Dry
+  module Schema
+    module Macros
+      Hash.prepend(::Module.new {
+        def call(*args)
+          if args.size >= 1 && args[0].is_a?(::Class) && args[0] <= ::Dry::Struct
+            if block_given?
+              raise ArgumentError, 'blocks are not supported'
```
solnic:
```suggestion
              raise ArgumentError, "blocks are not supported when using a struct class (#{name} => #{args[0]})"
```

flash-gordon:
@solnic it's something insane like `value(StructOne, StructTwo)` or 
```ruby
required(:foo).value(User) do
  require(:i_have_no_idea_what_im_doing).value(:string)
end
```

solnic:
@flash-gordon 😆 then we're fine

PR #254, opened 2020-03-01
Consistently use Template#key when evaluating templates in backends

robhanlon22:
When I reintroduced the Template class, I added both the key and text of the message that was looked up at Template instantiation as attributes of the Template class. This led to an inconsistency in how message parameter interpolation is implemented—while the I18n backend just uses the key and looks up the message again, the YAML backend would use the text. This PR makes the approach consistent, and has both message backends look up the message again at the time that interpolation occurs.

### DIFF
--- a/lib/dry/schema/messages/template.rb
+++ b/lib/dry/schema/messages/template.rb
@@ -18,15 +18,15 @@ module Dry
         option :options
 
         # @api private
-        def data(input = EMPTY_HASH)
+        def data(data = EMPTY_HASH)
           ensure_message!
-          messages.interpolatable_data(self, **options, **input)
+          messages.interpolatable_data(key, options, **options, **data)
         end
 
         # @api private
         def call(data = EMPTY_HASH)
           ensure_message!
-          messages.interpolate(self, **data)
+          messages.interpolate(key, options, **data)
         end
         alias_method :[], :call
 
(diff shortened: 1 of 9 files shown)


lib/dry/schema/messages/yaml.rb
```diff
@@ -146,31 +141,49 @@ def prepare
...
       private
 
-      def evaluation_context(template)
-        cache.fetch_or_store(template.text) do
-          tokens = template.text.scan(TOKEN_REGEXP).flatten(1).map(&:to_sym).to_set
-          text = template.text.gsub('%', '#')
+      # @api private
+      def evaluate_template(template)
+        result = get(template.key, template.options)
+        return unless result
```
robhanlon22: @solnic do you think this should `raise`?

lib/dry/schema/messages/template.rb
```diff
@@ -10,15 +11,15 @@ module Messages
...
 
         option :messages
         option :key
-        option :text
         option :options
 
         # @api private
         def data(input = EMPTY_HASH)
-          messages.pruned_data(self, **options, **input)
+          messages.interpolatable_data(self, **options, **input)
```
robhanlon22: I thought that this name (while long) made things more self-documentaing

lib/dry/schema/messages/yaml.rb
robhanlon22: Decided to raise a `KeyError` instead of returning `nil`, because this should only happen in exceptional cases

solnic:
@robhanlon22 it's not immediately obvious to me what this PR fixes/improves - could you summarize this somehow please?r

robhanlon22:
@solnic clarified

PR #255, opened 2020-03-02
Bump RuboCop, --safe-auto-correct the whole codebase

robhanlon22:
Totally understandable if this is not desirable, but I thought it couldn't hurt to open a PR!

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,4 +1,3 @@
-#!/usr/bin/env rake
 # frozen_string_literal: true
 
 require 'bundler/gem_tasks'
--- a/Gemfile
+++ b/Gemfile
@@ -6,9 +6,9 @@ eval_gemfile 'Gemfile.devtools'
 
 gemspec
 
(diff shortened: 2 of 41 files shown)


.rubocop.yml
```diff
@@ -1,7 +1,7 @@
 # this file is managed by dry-rb/devtools project
 
 AllCops:
-  TargetRubyVersion: 2.4
+  TargetRubyVersion: "2.7"
```
robhanlon22: This was needed as pattern matching support was added

flash-gordon: This file comes from devtools

dry-schema.gemspec
```diff
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 # this file is managed by dry-rb/devtools project
```
flash-gordon: This file is automatically generated from project.yml

.rubocop.yml
robhanlon22: oops! thanks

Rakefile
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/env rake
```
robhanlon22: Removing this seemed more sensible than giving the Rakefile execute permission

PR #260, opened 2020-03-07
Unexpected keys errors

### LINKED ISSUE #35: Generating errors for unexpected keys in Hash
I would love to be able to define a schema which generated errors if it was given unexpected keys:

``` ruby
require 'dry/validation'

schema = Dry::Validation.Schema(strict: true) do
  hash? do
    required(:foo).filled(:int?)
    required(:bar).filled(:str?)
  end
end

outcome = schema.call(foo: 1, bar: "two", baz: 3.0)
p outcome.messages # => {:baz=>["is not allowed"]}
```

This is really important for defining a public JSON API where you want to reject invalid user input.

solnic:
This adds a new processor step called `:key_validator` which validates if there are no unexpected keys. This works with both nested hashes and arrays with hashes.

```ruby
require 'dry/schema'

UserSchema = Dry::Schema.Params do
  config.validate_keys = true

  required(:address).hash do
    required(:city).filled(:string)
    required(:zipcode).filled(:string)
  end

  required(:roles).array(:hash) do
  end
end

input = {
  foo: 'unexpected',
  name: 'Jane',
  address: { bar: 'unexpected', city: 'NYC', zipcode: '1234' },
  roles: [{ name: 'admin' }, { name: 'editor', foo: 'unexpected' }]
}

UserSchema.(input).errors.to_h
# {
#  :foo=>["is not allowed"],
#  :address=>{:bar=>["is not allowed"]},
#  :roles=>{1=>{:foo=>["is not allowed"]}}
# }
```

Closes #35 

[changelog]

added: "You can now validate keys by setting `config.validate_keys = true". When enabled, schemas will provide error messages about unexpected keys in the input (issue #35 closed via #260) (@solnic)"

### DIFF
--- a/lib/dry/schema/key_validator.rb
+++ b/lib/dry/schema/key_validator.rb
@@ -1,0 +1,66 @@
+# frozen_string_literal: true
+
+require 'dry/initializer'
+require 'dry/schema/constants'
+
+module Dry
+  module Schema
+    # @api private
+    class KeyValidator
+      extend Dry::Initializer
+
+      INDEX_REGEX = /\[\d+\]/
+      DIGIT_REGEX = /\A\d+\z/
+      BRACKETS = '[]'
+
+      # @api private
+      option :key_map
+
+      # @api private
+      def call(result)
+        input = result.to_h
+
+        input_paths = key_paths(input)
+        key_paths = key_map.to_dot_notation
+
+        input_paths.each do |path|
+          error_path =
+            if path[INDEX_REGEX]
+              key = path.gsub(INDEX_REGEX, BRACKETS)
+
+              unless key_paths.include?(key)
+                arr = path.gsub(INDEX_REGEX) { |m| ".#{m[1]}" }
+                arr.split(DOT).map { |s| DIGIT_REGEX.match?(s) ? s.to_i : s.to_sym }
+              end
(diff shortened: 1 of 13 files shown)


lib/dry/schema/key_validator.rb
```diff
@@ -0,0 +1,58 @@
...
+        key_paths = key_map.to_dot_notation
+
+        input_paths.each do |path|
+          key = path.gsub(%r[\[\d+\]], '[]')
+          arr = path.gsub(/\[\d+\]/) { |m| ".#{m[1]}" }
+
+          unless key_paths.include?(key)
+            result.add_error(
+              [:unexpected_key, [
+                arr.split(DOT).map { |s| s =~ /\d+/ ? s.to_i : s.to_sym },
```
flash-gordon: should be `/\A\d+\z/` _I think_

lib/dry/schema/key_validator.rb
```diff
@@ -0,0 +1,58 @@
...
+        end
+
+        result
+      end
+
+      private
+
+      # @api private
+      def key_paths(hash)
+        hash.keys.flat_map { |key|
```
flash-gordon: `hash.flat_map { |key, value|`

PR #263, opened 2020-03-08
Fix setting type in nested schemas when a custom type was used

### LINKED ISSUE #174: Params type conversion is ignored
**Describe the bug**

**To Reproduce**

```
# Custom type
module CustomTypes
  include Dry.Types()

  Person = Nominal::Any.constructor(&:to_h)
end

# Person class
class Person
  def to_h
    { name: 'Luke', age: 21 }
  end
end

# Person schema
PersonSchema = Dry::Schema.Params do
  required(:age).value(:integer, gt?: 18)
end

# Contract
class NewUserContract < Dry::Validation::Contract
  params do
    required(:person).type(CustomTypes::Person).schema(PersonSchema)
  end
end
```

Then, instead of converting params using custom type `Person`, the error is being raised:

```
luke = Person.new

contract = NewUserContract.new

contract.call(person: luke)

undefined method `key?' for #<Person:0x00007fdd48873390>
```

Converting object to hash explicitly works, but then the `type(Type)` portion of the contract / schema definition is useless.

```
luke = Person.new

contract = NewUserContract.new

contract.call(person: luke.to_h)

#<Dry::Validation::Result{:person=>{:name=>"Luke", :age=>21}} errors={}>
```

**Expected behavior**

```
luke = Person.new

contract = NewUserContract.new

contract.call(person: luke)

#<Dry::Validation::Result{:person=>{:name=>"Luke", :age=>21}} errors={}>
```

**Your environment**

- Affects my production application: Yes
- Ruby version: 2.6.3
- OS: 10.14.6

solnic:
Fixes #174 

[changelog]

fixed: "Custom type is respected when defining nested schemas (issue #174 closed via #263) (@solnic)"

### DIFF
--- a/lib/dry/schema/dsl.rb
+++ b/lib/dry/schema/dsl.rb
@@ -176,7 +176,7 @@ module Dry
       def key(name, macro:, &block)
         raise ArgumentError, "Key +#{name}+ is not a symbol" unless name.is_a?(::Symbol)
 
-        set_type(name, Types::Any)
+        set_type(name, Types::Any.meta(default: true))
 
         macro = macro.new(
           name: name,
@@ -311,6 +311,15 @@ module Dry
         meta = { required: false, maybe: type.optional? }
 
         types[name] = type.meta(meta)
+      end
+
+      # Check if a custom type was set under provided key name
+      #
+      # @return [Bool]
+      #
+      # @api private
+      def custom_type?(name)
+        !types[name].meta[:default].equal?(true)
       end
 
       # Resolve type object from the provided spec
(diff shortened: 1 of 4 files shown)


lib/dry/schema/dsl.rb
```diff
@@ -176,7 +176,7 @@ def optional(name, &block)
       def key(name, macro:, &block)
         raise ArgumentError, "Key +#{name}+ is not a symbol" unless name.is_a?(::Symbol)
 
-        set_type(name, Types::Any)
+        set_type(name, Types::Any.meta(default: true))
```
flash-gordon: Does it make more sense to use a constant `DefaultType = Types::Any.dup.freeze` and then compare with `equal?`?

solnic: @flash-gordon it actually doesn't work in some cases, so I used meta instead, which is carried over in these cases and so this check works.

PR #267, opened 2020-03-09
Bump dry-types version

### LINKED ISSUE #229: JSON schema coercing empty string to nil
Using the json schema, empty strings are being coerced to nil.

 ```ruby
class MyContact < Dry::Validation::Contract
  json do
    optional(:value).maybe(:string)
  end
end

MyContract.new.call({ value: '' })
#<Dry::Validation::Result{:value=>nil} errors={}>
```

Since this is an optional maybe string, it should not coerce to nil.

    Affects my production application: NO
    Ruby version: 2.6.2
    OS: MacOS 10.15.2

flash-gordon:
Also add spec ensuring empty strings are not coerced to nil in JSON schemas.

### DIFF
--- a/project.yml
+++ b/project.yml
@@ -20,4 +20,4 @@ gemspec:
     - [dry-equalizer, "~> 0.2"]
     - [dry-initializer, "~> 3.0"]
     - [dry-logic, "~> 1.0"]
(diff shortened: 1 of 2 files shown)


flash-gordon:
I'll update changelog in master :)

PR #273, opened 2020-03-14
Make type specs mandatory

### LINKED ISSUE #272: Make type specs mandatory
Currently specifying types in the DSL is not mandatory and by default, whenever you specify a key via `required` or `optional` methods, [a default `Types::Any` is set](https://github.com/dry-rb/dry-schema/blob/master/lib/dry/schema/dsl.rb#L180).

This was done like that just to make it simpler to upgrade from dry-validation `0.x` to `1.0` where dry-schema is used as the schema backend. Unfortunately having optional type specs introduced a lot of complexity in the DSL itself.

In dry-schema 2.0.0 type specs should become mandatory. This means that every DSL method that doesn't have a type predefined should expect a type spec as the first argument.

## Examples

```ruby
# 1.x
required(:name).value
required(:name).filled

# 2.x
required(:name).value(:string)
```

robhanlon22:
Resolves #272 

- [ ] Fix all of the specs
- [ ] Fix Macros::Array#value to correctly coerce hash types
- [ ] Should filter be usable without a type spec?

### DIFF
--- a/lib/dry/schema/macros/array.rb
+++ b/lib/dry/schema/macros/array.rb
@@ -10,32 +10,30 @@ module Dry
       # @api private
       class Array < DSL
         # @api private
-        def value(*args, **opts, &block)
+        def value(*predicates, type_spec:, type_rule:, **opts, &block)
           type(:array)
 
-          extract_type_spec(*args, set_type: false) do |*predicates, type_spec:, type_rule:|
-            type(schema_dsl.array[type_spec]) if type_spec
+          type(schema_dsl.array[type_spec]) if type_spec
 
-            is_hash_block = type_spec.equal?(:hash)
+          is_hash_block = type_spec.equal?(:hash)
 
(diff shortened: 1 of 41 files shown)


lib/dry/schema/macros/array.rb
```diff
@@ -10,10 +10,10 @@ module Macros
...
         # @api private
-        def value(*args, **opts, &block)
+        def value(type_spec, *args, **opts, &block)
           type(:array)
 
-          extract_type_spec(*args, set_type: false) do |*predicates, type_spec:, type_rule:|
+          extract_type_spec(type_spec, *args, set_type: false) do |*predicates, type_spec:, type_rule:|
             type(schema_dsl.array[type_spec]) if type_spec
 
             is_hash_block = type_spec.equal?(:hash)
```
robhanlon22: This is quite tricky to handle, since this isn't always going to be true—it's possible to have a hash type_spec that takes no block.

PR #284, opened 2020-04-11
Fix key validation when working with arrays

### LINKED ISSUE #283: Unexpected NoMethodError when key validation is true
**Describe the bug**

When requiring that an array has string values, and also setting `config.validate_keys = true`, calling the schema raises a surprising NoMethodError. The error message that says `flat_map` cannot be called on the first string element in the given array.

This seems related to #281.

**To Reproduce**

Add `config.validate_keys = true` to the example on [this dry.rb page][page].

```ruby
  config.validate_keys = true
  
  optional(:tags).array(:str?)
end

result = schema.call(name: "", tags: ["red", 123])

#=> NoMethodError: undefined method `flat_map' for "red":String
#=> from /Users/cloyd/code/aspen/aspen/vendor/bundle/ruby/2.6.0/gems/dry-schema-1.5.0/lib/dry/schema/key_validator.rb:51:in `key_paths'
```

**Expected behavior**

Given that no unexpected keys are present, I expected this example to produce the same exact results on the [aforementioned page][page]. That is, it should have validated, not throwing a NoMethodError, with some resulting validation error messages.

**Your environment**

- Ruby version: 2.6.5
- OS: macOS Mojave 10.14.6

[page]: https://dry-rb.org/gems/dry-schema/1.5/basics/working-with-schemas/

beechnut:
Fixes #283.

When both requiring an array attribute and validating keys, dry-schema would throw a NoMethodError when validating keys.

This adds a check to guard against validating keys inside Arrays that just contain non-hash values.

### DIFF
--- a/spec/integration/schema/unexpected_keys_spec.rb
+++ b/spec/integration/schema/unexpected_keys_spec.rb
@@ -44,7 +44,7 @@ RSpec.describe Dry::Schema, "unexpected keys" do
         config.validate_keys = true
 
         required(:name).filled(:string)
-        optional(:tags).array(:str?)
+        optional(:tags).array(:string)
       end
     end
 
--- a/spec/integration/schema/unexpected_keys_spec.rb
+++ b/spec/integration/schema/unexpected_keys_spec.rb
@@ -37,4 +37,25 @@ RSpec.describe Dry::Schema, "unexpected keys" do
         roles: {1 => {foo: ["is not allowed"]}}
       )
   end
+
+  context "with an array validation" do
+    subject(:schema) do
+      Dry::Schema.define do
+        config.validate_keys = true
+
+        required(:name).filled(:string)
+        optional(:tags).array(:str?)
+      end
+    end
+
+    it "adds error messages" do
+      input = { name: "", tags: ["red", 123] }
+      expect(schema.(input).errors.to_h)
+        .to eql(
+          name: ["must be filled"],
+          tags: {1=>["must be a string"]}
+        )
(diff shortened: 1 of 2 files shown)


solnic:
Could you rebase on top of latest mater? I had to downgrade ruby 2.7.1 to 2.7.0 in the matrix to make things pass (we found a bug in 2.7.1 that causes segfaults).

beechnut:
Alright, the rebase helped with the 2.7 errors, tackling the 2.4 failures next.

beechnut:
@solnic When I click on the Details link for one of the checks, it shows all the checks as passing, but here it looks like it's waiting for some additional 2.7.* checks. I'm not sure how to resolve this, but please let me know if there's anything you think I should do.

spec/integration/schema/unexpected_keys_spec.rb
```diff
@@ -37,4 +37,25 @@
...
       )
   end
+
+  context "with an array validation" do
+    subject(:schema) do
+      Dry::Schema.define do
+        config.validate_keys = true
+
+        required(:name).filled(:string)
+        optional(:tags).array(:str?)
```
solnic, line 47:
```suggestion
        optional(:tags).array(:string)
```

beechnut:
@solnic Changes made! I'm seeing the same issue with CI from before: 2.7.0 is passing, and 2.7(.*) is hanging.

solnic:
@beechnut it's merged, thanks!

Issue #290, opened 2020-03-18
Changed interpolated string argument in error message


tadeusz-niemiec:
Hey, after upgrading to 1.5.0 I found that interpolated argument for error in `size?:` key is now called `size` but before 1.5.0 it was a `num`.

Example from `pl.yml` config file:

This worked before and now doesn't:
`size?: "musi mieć %{num} znaków"`
and produces an error:
`missing interpolation argument :num in "musi mieć %{num} znaków"`

This works now:
`size?: "musi mieć %{size} znaków"`

I'd like to mention also that the interpolated argument name remained unchanged for `max_size?:` and `min_size?:` error keys - it's still called "num"

EDIT:

This is actually caused by inconsistency in argument names in the dry-logic methods. "min_size?" and "max_size?" arguments are called "num", but the "size?" method has the argument called "size". Dry schema calls `predicate.parameters` and uses them as interpolated argument names in error messages. Actually tried to fix this by simply changing the argument name, but it broke a lot of stuff in the dry-schema :(

ref: https://github.com/dry-rb/dry-logic/pull/72

PR #291, opened 2020-05-27
Fix for inconsistency in the interpolated arguments names

tadeusz-niemiec:
Fixes dry-rb/dry-schema#290.

Now all "size" validations can handle both "num" and "size" interpolated args.

### DIFF
--- a/lib/dry/schema/message_compiler.rb
+++ b/lib/dry/schema/message_compiler.rb
@@ -216,7 +216,7 @@ module Dry
 
       # @api private
       def message_tokens(args)
-        args.each_with_object({}) do |arg, hash|
+        tokens = args.each_with_object({}) do |arg, hash|
           case arg[1]
           when Array
             hash[arg[0]] = arg[1].join(LIST_SEPARATOR)
@@ -227,6 +227,14 @@ module Dry
             hash[arg[0]] = arg[1]
           end
         end
+        args.any? { |e| e.first == :size } ? append_mapped_size_tokens(tokens) : tokens
+      end
+
+      # @api private
+      def append_mapped_size_tokens(tokens)
+        # this is a temporary fix for the inconsistency in the "size" errors arguments
+        mapped_hash = tokens.each_with_object({}) { |(k, v), h| h[k.to_s.gsub("size", "num").to_sym] = v }
+        tokens.merge(mapped_hash)
       end
     end
   end
--- a/spec/integration/message_compiler_spec.rb
+++ b/spec/integration/message_compiler_spec.rb
@@ -14,7 +14,7 @@ RSpec.describe Dry::Schema::MessageCompiler do
           errors: {
             key?: {
               arg: {
-                default: "+%{name}+ key is missing in the hash"
+                default: '+%{name}+ key is missing in the hash'
               },
               value: {
                 gender: "Please provide your gender"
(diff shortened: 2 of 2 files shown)


solnic:
I'm afraid we can't change it like that - changing keys or tokens in errors.yml *is always a breaking change*. This would require a major version bump and I don't think it's worth it.

I gotta be honest that I don't understand what exactly broke it in the first place *but* it can be probably addressed by tweaking `MessageCompiler` so that it renames `size` to `num` in case of the `size?` predicate.

Then in 2.0.0 we could clean it up and rename the arg on the dry-logic side and remove the `size => num` mapping from `MessageCompiler`.

Does it make sense?

tadeusz-niemiec:
@solnic If I understand you correctly you want to pass the "size" argument all the way up to the MessageCompiler and then map it to "num" there, but imho this would also require writing errors.yml messages using "num" argument instead of "size" in the error message since we would be passing it as the interpolated argument.

The other possibility is that you meant to leave default errors.yml untouched and map this value only when the localization is taking place?

solnic:
> The other possibility is that you meant to leave default errors.yml untouched and map this value only when the localization is taking place?

Yes the goal is to not change errors.yml and restore the original behavior by tweaking MessageCompiler somehow. I just don't know how exactly because I didn't work on the change that caused the regression and I don't have time to dig into it 😢

tadeusz-niemiec:
@solnic please let me know what you think about this solution. I merge a new hash with mapped tokens to the original tokens hash. This way we are able to use "num" and "size" arguments interchangeably.

lib/dry/schema/message_compiler.rb
```diff
@@ -199,6 +199,14 @@ def message_tokens(args)
             hash[arg[0]] = arg[1]
           end
         end
+        append_mapped_size_tokens(tokens)
```
solnic, line 202: @tadeusz-niemiec could you make it so that this is used only when args include `size` key? Remapping tokens even when it's not needed will have a performance impact

tadeusz-niemiec: @solnic done

lib/dry/schema/message_compiler.rb
```diff
@@ -199,6 +199,14 @@ def message_tokens(args)
             hash[arg[0]] = arg[1]
           end
         end
+        args.any? { |e| e.first == :size } ? append_mapped_size_tokens(tokens) : tokens
+      end
+
+      # @api private
+      def append_mapped_size_tokens(tokens)
+        # this is a temporary fix for the inconsistency in the "size" errors arguments
+        mapped_hash = tokens.inject({}) { |h, (k, v)| h[k.to_s.gsub("size", "num").to_sym] = v; h }
```
robhanlon22, line 208: I _think_ this would be a good place to use `transform_keys` if I'm not crossing any wires on the logic here.

```suggestion
        tokens.transform_keys do |key|
          k.to_s.gsub("size", "num").to_sym
        end
```

spec/integration/message_compiler_spec.rb
```diff
@@ -316,6 +316,34 @@
 
         expect(msg).to eql('length must be within 3 - 4')
       end
+
+      it "handles 'num' and 'size' interchangeably as interpolated arguments" do
+        messages = Dry::Schema::Messages::YAML.build.merge(
```
robhanlon22, line 321: Would you mind running RuboCop on this file to make the indentation consistent?

solnic:
> (Also @solnic sorry for falling off the face of the earth for a while there after the 1.5.0 release)

@robhanlon22 no worries!

lib/dry/schema/message_compiler.rb
tadeusz-niemiec: Done, rubocop also removed a comma in one hash, but I guess it's ok

spec/integration/message_compiler_spec.rb
tadeusz-niemiec: Haha, that was even a method which was there in a first place but unfortunately I guess it was introduced after Ruby 2.4 and some of CLI for older version were failing with NoMethodErrors :(
I changed it to the `each_with_object`, I guess it's a little bit less awkward

lib/dry/schema/message_compiler.rb
tadeusz-niemiec: That was even a method which was there in a first place, but unfortunately it was introduced in Ruby 2.5 and CLIs for Ruby 2.4 were failing with NoMethodErrors :(

I changed it to the `each_with_object`, I guess it's a little bit less awkward

tadeusz-niemiec:
> Could you rebase/squash commits before I can merge this in?

Done :)

Issue #161, opened 2019-06-12
Retrieving errors with option full: true adds whitespace even in locales that don't have whitespace between words


cmcolinh:
Here is some example code that reproduces this issue in 1.0.0
```
class MyContract < Dry::Validation::Contract
  config.messages.backend        = :i18n
  config.messages.default_locale = :ja
  params do
    required(:name).filled(:string)
  end
  rule(:name){key.failure(:oh_no) if values[:name].eql?('oh no')}
end
```
with ja.yml
```
ja:
  dry_validation:
    errors:
      filled?: "は必須入力です"
      oh_no:   "はオーノーです"
    rules:
      name:    "名称"
```
Invoke with(error in params)
```
MyContract::new.(name: '').errors(full: true).to_h
```
expected_result: {:name=>["名称は必須入力です"]} (no whitespace between '名称' and 'は')
actual_result: {:name=>["名称 は必須入力です"]} (whitespace inserted)

Invoke with(error in rule)
```
MyContract::new.(name: 'oh no').errors(full: true).to_h
```
expected_result: {:name=>["名称はオーノーです"]} (no whitespace between '名称' and 'は')
actual_result: {:name=>["名称 はオーノーです"]} (whitespace inserted)

solnic:
I moved this over here and marked it as a feature, because it's essentially a missing ability to configure behavior of "full messages" (whether a space should be used or not).

cmcolinh:
I was sort of suspecting it would be as such, the best way to implement this does not immediately come to mind. Possibly by adding some key to the config object perhaps.  I could see something perhaps like
```
config.full_messages_whitespace[:ja] = false
```
But even that doesn't seem so self-documenting (also, that would be unwieldy in the source code in question)
e.g. (I'm not sure I have this right at all)
```
"#{messages.rule(rule, options) || rule}#{' ' unless config.full_messages_whitespace[options.locale].eql?(false)}#{text}"
```

solnic:
We just need handlers for each locale identifier, with a default one that does what we already have, and the special one for `:ja`. Then we can do:

```ruby
full_text_handlers[locale].call(rule, options)
```

This can be as simple as a hash with a default value set.

cmcolinh:
Would the handler be passed in as an injectable dependency?

e.g
[...]

PR #292, opened 2020-05-27
Allow customization of whitespace character between words per locale

tadeusz-niemiec:
Resolves dry-rb/dry-schema#161

This allows to define the whitespace char between words in localized full error messages.


solnic:
Thank you! You're on 🔥 😄

Issue #301, opened 2020-07-07
Schema.JSON documentation wrong


juriglx:
Executing the example code provided at https://dry-rb.org/gems/dry-schema/1.4/json/ results in different behaviour than claimed.

I expect to get:
`:age => ["must be greater than 18"]`

I get:
`:age => ["must be an integer"]`

I tested this in isolation with dry-schema 1.4.0

PR #302, opened 2020-07-07
Fix json documentation

juriglx:
See Issue https://github.com/dry-rb/dry-schema/issues/301


solnic:
hey, thanks for spotting this and opening this PR. I fixed it via #303 after all.

PR #305, opened 2020-07-22
Avoid creation of array in Info::SchemaCompiler.visit_set

esparta:
The goal of `visit_set` is visit the nodes using their target, but while it does it also creates an array is not used on the method.


esparta:
Thanks!

Issue #311, opened 2020-08-28
"is not allowed" raised when `validate_keys` is enabled for maybe hashes


svobom57:
**Describe the bug**
There was an issue: https://github.com/dry-rb/dry-schema/issues/308 whereby the validation of keys for nested maybe hashes weren't working prior to version `1.5.3`. There's still an issue with non-nested maybe hashes.

**To Reproduce**
```ruby
schema = Dry::Schema.Params do
  config.validate_keys = true

  required(:feedback_location).maybe(:hash) do
    required(:lat).filled(:float)
    required(:lon).filled(:float)
  end
end

schema.call({ feedback_location: nil }).errors
=> #<Dry::Schema::MessageSet messages=[#<Dry::Schema::Message text="is not allowed" path="feedback_location" predicate=nil input={:feedback_location=>nil}>] options={:failures=>true}>
```

**Expected behaviour**

It shouldn't have an error since the `feedback_location` field is a `maybe` hash.

**Your environment**

- Affects my production application: **YES**
- Ruby version: 2.7
- dry-schema (1.5.3)
dry-validation (1.5.3)
- OS: MacOS

svobom57:
@solnic When are you planning on cutting the `1.5.4` release?

solnic:
Tomorrow

PR #312, opened 2020-08-28
 Fix bug: non-nested maybe hashes have errors on `validate_keys = true`

svobom57:
### Description
There is an issue whereby if we're validating keys on `maybe` hashes in non-nested strucutures the schema comes back with `is not allowed`:

```ruby
  config.validate_keys = true

  required(:feedback_location).maybe(:hash) do
    required(:lat).filled(:float)
    required(:lon).filled(:float)
  end
end

schema.call({ feedback_location: nil }).errors
=> #<Dry::Schema::MessageSet messages=[#<Dry::Schema::Message text="is not allowed" path="feedback_location" predicate=nil input={:feedback_location=>nil}>] options={:failures=>true}>
```

This mitigates the error for checking the partial path similarly to what has been done for the nested structures here:
https://github.com/dry-rb/dry-schema/pull/309

### Context

Closes: https://github.com/dry-rb/dry-schema/issues/311

/cc @solnic

### DIFF
--- a/lib/dry/schema/key_validator.rb
+++ b/lib/dry/schema/key_validator.rb
@@ -32,7 +32,7 @@ module Dry
                 arr = path.gsub(INDEX_REGEX) { |m| ".#{m[1]}" }
                 arr.split(DOT).map { |s| DIGIT_REGEX.match?(s) ? s.to_i : s.to_sym }
               end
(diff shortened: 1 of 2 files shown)


solnic (approved):
👏 🙇

PR #313, opened 2020-09-03
support meta tags in unexpected key error messages

### LINKED ISSUE #310: KeyValidator doesn't support text/code meta data in messages
Hello,

**Describe the bug**

I need error messages with codes and texts. So I use the 'meta' feature in my messages files (hashes with `text` and `code` keys).
I activate the KeyValidator option:  `config.validate_keys = true`.
I have 2 formatting issues:
* The produced error does not have a structure as hash with the `code`  and  `text` keys.
* The produced error does not seem to use the `full: true` option to add the key into the error text.

**To Reproduce**

````
# my_messages.yml

en:
  dry_validation_with_codes:
    or: "or"

    errors:
      unexpected_key:
        text: "is not allowed"
        code: "is_not_allowed"
      key?:
        text: "is missing"
        code: "missing"
````

````
# my_contract.rb

require 'dry/validation/contract'

class MyContract < Dry::Validation::Contract
  # Define a top namespace and use only my messages file in loaded paths
  # to avoid conflict with default messages from the gem.
  config.messages.top_namespace = 'dry_validation_with_codes'
  config.messages.load_paths = ['my_messages.yml']
  config.validate_keys = true

  params do
    required(:title).filled
  end
end


result = MyContract.new.call({})
puts result.errors(full: true).to_h.inspect
# => {:title=>[{:text=>"title is missing", :code=>"missing"}]}
# OK

result = MyContract.new.call({badkey: "value"})
puts result.errors(full: true).to_h.inspect

# => {:badkey=>["is not allowed"], :title=>[{:text=>"title is missing", :code=>"missing"}]}
# Not OK, expected result:
# {:badkey=>[{:text=>"badkey is not allowed" :code=>"is_not_allowed"}]}, :title=>[{:text=>"title is missing", :code=>"missing"}]}
````

**Expected behavior**

Expected result for KeyValidator message: 
`{:badkey=>[{:text=>"badkey is not allowed", :code=>"is_not_allowed"}]}`

Observed result: 
`{:badkey=>["is not allowed"]}`

**Your environment**

- Ruby version: ruby 2.7.1p83
- Gem dry-validation: 1.5.4
- OS: FreeBSD 12.1

Thank you.

### PR DESCRIPTION
**Pull request by tadeusz-niemiec**
Fixes #310

### DIFF
--- a/lib/dry/schema/message_compiler.rb
+++ b/lib/dry/schema/message_compiler.rb
@@ -116,6 +116,7 @@ module Dry
 
         Message.new(
           path: path,
+          meta: msg[:meta] || EMPTY_HASH,
           text: msg[:text],
           predicate: nil,
           input: input
--- a/spec/fixtures/locales/pl.yml
+++ b/spec/fixtures/locales/pl.yml
@@ -23,3 +23,9 @@ pl:
         rules:
           email:
             filled?: "Hej user! Dawaj ten email no!"
+      unexpected_key:
+        text: "Podano nieoczekiwany klucz"
+        code: "nieoczekiwany_klucz"
+      key?:
+        text: "nie został podany"
+        code: "brak_klucza"
--- a/spec/integration/schema/unexpected_keys_spec.rb
+++ b/spec/integration/schema/unexpected_keys_spec.rb
@@ -55,6 +55,20 @@ RSpec.describe Dry::Schema, "unexpected keys" do
     expect(schema.(input)).to be_failure
   end
 
+  it "supports meta tags" do
+    schema = Dry::Schema.define do
+      config.validate_keys = true
+      config.messages.load_paths << "#{SPEC_ROOT}/fixtures/locales/pl.yml"
+      config.messages.default_locale = :pl
+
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: lib/dry/schema/message_compiler.rb [hunk 1]
```diff
@@ -116,6 +116,7 @@ def visit_unexpected_key(node, _opts)
 
         Message.new(
           path: path,
+          meta: msg[:meta] || {},
```
solnic, line 119:
```suggestion
          meta: msg[:meta] || EMPTY_HASH,
```

This way we won't be allocating extra hash. Message objects are "immutable" so this is perfectly fine.

solnic (changes requested):
Thanks for taking care of this. There's just one small issue to address before I can merge this in. Please also rebase after applying the suggestion! 🙇

tadeusz-niemiec:
> Thanks for taking care of this. There's just one small issue to address before I can merge this in. Please also rebase after applying the suggestion!

Done :)

solnic (approved):
👏 👏 👏 🙇

PR #315, opened 2020-09-07
Fix errors when hash keys are provided as strings

### LINKED ISSUE #296: Nested optional params schemas don't accept string keys.
When there's an optional hash/schema value for a params key, the expectation is that the nested hash's keys are symbols. Going by @solnic's comments in [this discussion](https://discourse.dry-rb.org/t/matching-optional-params-schema-against-a-nested-schema/1069/5), it sounds like a bug (and with my limited knowledge, I agree - especially given params data via HTTP requests is always going to be strings).

```ruby
require "dry/schema"

inner_schema = Dry::Schema.Params do
  required(:path).filled(:string)
end

outer_schema = Dry::Schema.Params do
  required(:foo).maybe(:hash, inner_schema)
end

# This works as expected
puts outer_schema.("foo" => {path: "bar"}).inspect
# => #<Dry::Schema::Result{:foo=>{:path=>"bar"}} errors={}>

# This fails :(
puts outer_schema.("foo" => {"path" => "bar"}).inspect
# => #<Dry::Schema::Result{:foo=>{"path"=>"bar"}} errors={:foo=>{:path=>["is missing"]}}>
```

From some quick testing, this only appears to be a bug for optional values (`maybe`). Using `hash(inner_schema)` and `filled(inner_schema)` both return no validation errors for both symbol and string keys.

My testing for this has been on macOS with MRI 2.6.6, I've not attempted it in production on Linux, but I doubt it's environment-specific.

### LINKED ISSUE #306: Composed schema incorrectly handle nested schemas
**Describe the bug**

It appears when using an `or`ed schema in an array, schemas nested below that will not pass validation when they should.

**To Reproduce**

```ruby
require "dry/validation"

ScaleOptionsSchema = Dry::Schema.Params do
  required(:value)
  required(:text).filled(:string)
end

ScaleSchema = Dry::Schema.Params do
  required(:type).filled(eql?: "scale")
  required(:id).filled(:string)
  required(:options)
    .value(:array, min_size?: 1)
    .each(:hash, ScaleOptionsSchema)
end

QuestionSchema = Dry::Schema.Params do
  required(:type).filled(eql?: "question")
  required(:id).filled(:string)
  required(:text).filled(:string)
end

Schema = Dry::Schema.Params do
  required(:definitions)
    .value(:array, min_size?: 1)
    .each(:hash, ScaleSchema | QuestionSchema)
end

valid_hash = {
  "definitions" => [
    {
      "type" => "scale",
      "id" => "1",
      "options" => [
        {
          "text" => "No",
          "value" => 1,
        },
        {
          "text" => "Yes",
          "value" => 2,
        },
      ],
    },
    {
      "id" => "2",
      "type" => "question",
      "text" => "hello",
    },
  ],
}

validator = Schema.call(valid_hash)
puts validator.success?
puts validator.errors.to_h.inspect # contains errors
```

**Expected behavior**

Expected no errors to be be found by the validator. This may not work until 2.0 based on the experimental nature of composing schemas (https://dry-rb.org/gems/dry-schema/1.5/advanced/composing-schemas/).

**Your environment**

- Ruby version: 2.6.5
- OS: Mac OS X 10.15.6

tadeusz-niemiec:
Closes #296.

### DIFF
--- a/spec/integration/schema/or_spec.rb
+++ b/spec/integration/schema/or_spec.rb
@@ -86,4 +86,29 @@ RSpec.describe Dry::Schema, "OR messages" do
       expect(schema.(foo: {bar: ""}).errors).to eql(foo: {bar: ["must be filled"]})
     end
   end
+
+  context "with two schemas" do
+    name_schema = Dry::Schema.define do
+      required(:name).filled(:string)
+    end
+
+    nickname_schema = Dry::Schema.define do
+      required(:nickname).filled(:string)
+    end
+
+    subject(:schema) do
+      Dry::Schema.define do
+        required(:user) { name_schema | nickname_schema }
+      end
+    end
+
+    it "returns success for valid input" do
+      expect(schema.(user: {name: "John"})).to be_success
+      expect(schema.(user: {nickname: "John"})).to be_success
+    end
+
+    it "provides error messages for invalid input where both sides failed" do
+      expect(schema.(user: {}).errors.to_h).to eql(user: {or: [{name: ["is missing"]}, {nickname: ["is missing"]}]})
+    end
+  end
 end
--- a/lib/dry/schema/macros/value.rb
+++ b/lib/dry/schema/macros/value.rb
@@ -28,10 +28,14 @@ module Dry
 
             import_steps(schema)
 
-            type(updated_type) unless custom_type? && !current_type.respond_to?(:of)
+            if !custom_type? || array_type?(current_type) || hash_type?(current_type)
+              type(updated_type)
+            elsif maybe_type?(current_type)
+              type(updated_type.optional)
+            end
           end
 
-          trace_opts = opts.reject { |key, _| key == :type_spec || key == :type_rule }
+          trace_opts = opts.reject { |key, _| %i[type_spec type_rule].include?(key) }
 
           if (type_rule = opts[:type_rule])
             trace.append(type_rule).evaluate(*predicates, **trace_opts)
@@ -62,8 +66,14 @@ module Dry
           primitive_inferrer[type].eql?([::Array])
         end
 
+        # @api private
         def hash_type?(type)
           primitive_inferrer[type].eql?([::Hash])
+        end
+
+        # @api private
+        def maybe_type?(type)
+          type.meta[:maybe].equal?(true)
         end
 
         # @api private
--- a/spec/integration/params/macros/maybe_spec.rb
+++ b/spec/integration/params/macros/maybe_spec.rb
@@ -77,4 +77,30 @@ RSpec.describe "Params / Macros / maybe" do
       expect(schema.(song: nil)).to be_success
     end
   end
+
+  context "with a nested schema" do
+    inner_schema = Dry::Schema.Params do
+      required(:name).filled(:string)
+    end
+
+    schema = Dry::Schema.Params do
+      required(:user).maybe(:hash, inner_schema)
+    end
+
+    it "passes when valid" do
+      expect(schema.("user" => {name: "John"})).to be_success
+    end
+
+    it "passes when valid and keys are strings" do
+      expect(schema.("user" => {"name" => "John"})).to be_success
+    end
+
+    it "fails when not valid" do
+      expect(schema.("user" => {"name" => 1}).errors.to_h).to eq(user: {name: ["must be a string"]})
+    end
+
+    it "passes when optional argument is missing" do
+      expect(schema.("user" => nil)).to be_success
+    end
+  end
 end
--- a/spec/integration/schema/macros/maybe_spec.rb
+++ b/spec/integration/schema/macros/maybe_spec.rb
@@ -139,4 +139,40 @@ RSpec.describe "Macros #maybe" do
       expect(schema.(song: nil)).to be_success
     end
   end
+
+  context "with a nested schema" do
+    inner_schema = Dry::Schema.define do
+      required(:name).filled(:string)
+    end
+
+    schema = Dry::Schema.define do
+      required(:user).maybe(:hash, inner_schema)
+    end
+
+    it "passes when valid" do
+      expect(schema.(user: {name: "John"})).to be_success
+    end
+
+    it "fails when not valid" do
+      expect(schema.(user: {name: 1}).errors.to_h).to eq(user: {name: ["must be a string"]})
+    end
+  end
+
+  context "with an array with nested schema" do
+    inner_schema = Dry::Schema.define do
+      required(:name).filled(:string)
+    end
+
+    schema = Dry::Schema.define do
+      required(:users).maybe(:array).each(inner_schema)
+    end
+
+    it "passes when valid" do
+      expect(schema.(users: [{name: "John"}])).to be_success
+    end
+
+    it "fails when not valid" do
+      expect(schema.(users: [{name: 1}]).errors.to_h).to eq(users: {0 => {name: ["must be a string"]}})
+    end
+  end
 end


lib/dry/schema/rule_applier.rb
```diff
@@ -32,7 +32,7 @@ def call(input)
         rules.each do |name, rule|
           next if input.error?(name)
 
-          result = rule.(input.to_h)
+          result = rule.(deep_symbolize_hash(input.to_h))
```
solnic, line 35: keys are meant to be coerced by the key map object, this is a separate step so here we expect the hash to already have symbol keys

lib/dry/schema/key.rb
```diff
@@ -46,6 +46,8 @@ def read(source)
           yield(source[name])
         elsif source.key?(coerced_name)
           yield(source[coerced_name])
+        elsif source.key?(name.to_s)
+          yield(source[name.to_s])
```
solnic, line 50: Why is this needed?

solnic (changes requested):
I don't entirely understand what the culprit is but this must be fixed at the key map level somehow :/

lib/dry/schema/key.rb
tadeusz-niemiec: This was a big fail, because I was testing a non-params schema by an accident, so all of my tests were failing when I passed any string key :facepalm:  

I'm currently refactoring it.

spec/integration/schema/or_spec.rb
```diff
@@ -86,4 +86,29 @@
       expect(schema.(foo: {bar: ""}).errors).to eql(foo: {bar: ["must be filled"]})
     end
   end
+
+  context "with two schemas" do
```
tadeusz-niemiec, line 90: This is not really related to this change, but I added a missing test case

tadeusz-niemiec:
> Codacy Here is an overview of what got changed by this pull request:
> 
> ```diff
> Clones added
> ============
> - spec/integration/schema/or_spec.rb  1
> - spec/integration/params/or_spec.rb  1
>          
> ```
> 
> See the complete overview on [Codacy](https://app.codacy.com/gh/dry-rb/dry-schema/pullRequest?prid=6138213&bid=20069217)

Ok, I actually copied the schema test case to test params schema and changed provided inputs to hashes with strings, should I remove/update it?

spec/integration/params/or_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+    it "provides OR message for invalid input where both sides failed" do
+      expect(schema.("foo" => []).errors).to eql(foo: ["must be a string or must be a hash"])
+    end
+
+    it "provides error messages for invalid input where right side rules failed" do
+      expect(schema.("foo" => {"bar" => ""}).errors).to eql(foo: {bar: ["must be filled"]})
+    end
+  end
+
+  context "with two schemas" do
```
tadeusz-niemiec, line 90: test which were failing without my changes start here

lib/dry/schema/key.rb
```diff
@@ -51,7 +51,7 @@ def read(source)
 
       # @api private
       def write(source, target)
-        read(source) { |value| target[coerced_name] = value }
+        read(source) { |value| target[coerced_name] = deep_symbolize_value(value) }
```
solnic, line 54: I'm sorry but I didn't explain things fully! The KeyMap is explicit, in the sense that it is built based on keys defined by a schema, and it uses the keys it knows about to sanitize an input hash. The whole point of this object is to sanitize a hash explicitly, rather than blindly coercing all the keys.

The problem here is that when a macro is used with `:hash` provided as the type and an external schema as the second arg, ie `value(:hash, SomeSchema)` then `SomeSchema` is not used to extend current's schema types, and key map *is built from types* so we end up with a key map that doesn't include `SomeSchema` keys.

Here's a fix that will probably make all your specs pass:

```diff
diff --git lib/dry/schema/macros/value.rb lib/dry/schema/macros/value.rb
index 905fd8c..665fba8 100644
--- lib/dry/schema/macros/value.rb
+++ lib/dry/schema/macros/value.rb
@@ -28,7 +28,9 @@ module Dry
 
             import_steps(schema)
 
-            type(updated_type) unless custom_type? && !current_type.respond_to?(:of)
+            if !custom_type? || array_type?(current_type) || hash_type?(current_type)
+              type(updated_type)
+            end
           end
 
           trace_opts = opts.reject { |key, _| key == :type_spec || key == :type_rule }
```

solnic (changes requested):

lib/dry/schema/key.rb
solnic: Correction: the above diff will fix only #296 

In case of #315 it's a bit more tricky as composed schemas are represented as dry logic operations, so we need a way to extract schema types from those in `Macros::Each#value`.

tadeusz-niemiec: Now I got it, thanks! 

I tried to implement your solution and it actually required one more change here:

```ruby
def hash_type?(type)
  primitive_inferrer[type].eql?([::Hash])
end
```

to:

```ruby
def hash_type?(type)
   primitive_inferrer[type].include?(::Hash)
end
```
The reason here is that `primitive_inferrer[current_type]` actually returns `[NilClass, Hash]`, instead of `[Hash]`
The above solution solves all problems with #296 and didn't break anything new, but I guess that the `updated_type` should be assigned only the proper side of `"NilClass | Hash"`, do I get it right?

Almost identical case happens with a schema nested under an array, like this:

```ruby
    inner_schema = Dry::Schema.define do
    end

      required(:users).maybe(:array).each(inner_schema)
    end

   schema.(users: [{name: "John"}]).errors.to_h
   >> {:users=>{0=>{:name=>["is missing"]}}}
```
and this fails for the similar reason, because `primitive_inferrer[current_type]` returns `[NilClass, Array]`, but this case is more tricky, since there is the whole `build_array_type` method.

Do you have any tips about these?

solnic: @tadeusz-niemiec ah great findings! In case of `maybe` I think we should enforce a maybe type at the `Macros::Maybe#call` level. It could check if the type that was set is already a maybe type, and if it's not - then override it via `type(current_type.optional)`. You could add a helper method to `Macros::Value` too called `maybe_type?` that would simply do `current_type.meta[:maybe].equal?(true)` and use that.

Makes sense?

tadeusz-niemiec: Makes sense and even works! ;)

So, the  #296 is fixed, but there are still problems with: 

1. An optional array with a nested schema (Honestly I don't have a good idea here and what's more - I can't test it properly, because there is a bug, which causes the optional array validation to fail when a `nil` is passed instead of an Array, like this: 

```ruby
     inner_schema = Dry::Schema.define do
    end

      required(:users).maybe(:array).each(inner_schema)
    end

   schema.(users: nil)
   >> NoMethodError: undefined method `map' for nil:NilClass
```

2.  A schema with composed two params schemas using OR (basically the #306)

To sum up - this PR is ready to be merged and it solves #296, but I also found a new issue (1st point) and the #306 is still unsolved. I can open a new issue if you'd like to.

solnic (approved):

lib/dry/schema/key.rb
solnic: @tadeusz-niemiec thanks for the hard work and sticking with me here :) Could you report an issue about the 1st point?

Re the second point, this is a somewhat difficult problem to solve. Composed schemas (using `OR`) mean that you may have either one set of keys or another in the input. This means `KeyMap` would have to be extended with the ability to check if a given set of keys exist and coerce that.

pat:
Thanks for this fix :)

Issue #317, opened 2020-10-06
`#error?(path)` doesn't seem to work correctly


pigats:
**Describe the bug**

`#error?(path)` doesn't seem to work correctly with nested validations

**To Reproduce**

```ruby
require "dry-schema"
require "json"
require "rspec"

class Schema
  BlockItem = Dry::Schema.JSON do
    optional(:payload).filled(:hash)
  end

  Item = Dry::Schema.JSON do
    required(:i18n).array(:hash) do
      required(:blocks).array(BlockItem)
    end
  end

  Items = Dry::Schema.JSON do
    required(:items).array(Item)
  end
end

payload = JSON.parse(
<<~JSON
  {
    "items": [
      {
        "i18n": [
          {
            "blocks": [
              {
                "payload": {
                  "key": "value"
                }
              },
              {
                "payload": {}
              }
            ]
          }
        ]
      },
      {
        "i18n": [
          {
            "blocks": [
              {
                "payload": {
                  "key": "value"
                }
              }
            ]
          }
        ]
      }
    ]
  }
JSON
)

RSpec.describe("#error?(path)") do
  it "is true, false" do
    result = Schema::Items.(payload)
    expect(result.error?([:items, 0])).to be true
    expect(result.error?([:items, 1])).to be false
  end
end
```

**Expected behavior**

`result.error?([:items, 1])` should be `false`, as the second item is valid.

**Notes**

Please note that changing the position of the invalid block affects the result - it seems the index passed to `#error?([:items, index])` is checking the block position rather than the item position - eg. with
 
```json
 {
    "items": [
      {
        "i18n": [
          {
            "blocks": [
              {
                "payload": {
                  "key": "value"
                }
              },              
              {
                "payload": {
                  "key": "value"
                }
              },
              {
                "payload": {}
              }
            ]
          }
        ]
      },
      {
        "i18n": [
          {
            "blocks": [
              {
                "payload": {
                  "key": "value"
                }
              }
            ]
          }
        ]
      }
    ]
  }
```

`result.error?([:items, 1])` is `false` and `result.error?([:items, 2])` is `true`, although there’s no 3rd item in the `items` array.

**Your environment**

- Affects my production application: Yes
- Ruby version:  2.6.5p114
- OS: CentOS Linux release 7.4.1708
- dry-schema: 1.5.4

solnic:
@pigats thanks for the detailed report Andrea <3 I'll fix it later today and push a release.

[...]

PR #319, opened 2020-10-08
Fix Result#error? when both sides include common parent array item

solnic:
Refs #317

### DIFF
--- a/lib/dry/schema/path.rb
+++ b/lib/dry/schema/path.rb
@@ -90,15 +90,29 @@ module Dry
         keys.index(key)
       end
 
+      def without_index
+        self.class.new([*to_a[0..-2]])
+      end
+
       # @api private
+      #
+      # rubocop:disable Metrics/PerceivedComplexity
       def include?(other)
         return false unless same_root?(other)
         return last.equal?(other.last) if index? && other.index?
-        return self.class.new([*to_a[0..-2]]).include?(other) if index?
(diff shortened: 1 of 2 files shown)


lib/dry/schema/path.rb
```diff
@@ -90,15 +90,29 @@ def index(key)
...
-        return false if !index? && other.index?
+        return without_index.include?(other) if index?
+
+        if !index? && other.index?
+          path = key_matches(other, :select)
+
+          return false unless path.size > 1
+
+          self.class.new(path).include?(other)
+        end
```
flash-gordon, line 111: Equivalent to this code
```ruby
      def include?(other)
        return false unless same_root?(other)
        return last.equal?(other.last) if index? && other.index?
        return without_index.include?(other) if index?
        return false if other.index? && key_matches(other, :select).size < 2

        self >= other && !other.key_matches(self).include?(nil)
      end
```
But I would prefer a form without `return`:
```ruby
      def include?(other)
        if !same_root?(other)
          false
        elsif index?
          if other.index?
            last.equal?(other.last)
          else
            without_index.include?(other)
          end
        elsif other.index? && key_matches(other, :select).size < 2
          false
        else
          self >= other && !other.key_matches(self).include?(nil)
        end
      end
```
because the latter doesn't have duplicated checks (see calls of `index?`) and IMHO a bit less prone to error

PR #332, opened 2020-11-13
Improve error message for attr? predicate

tylerhunt:
Changes the error message for the `attr?` predicate from:

```ruby
"is missing"
```
 to:
```ruby
"must respond to %{name}"
```

### DIFF
--- a/config/errors.yml
+++ b/config/errors.yml
@@ -67,7 +67,7 @@ en:
 
       key?: "is missing"
 
-      attr?: "is missing"
+      attr?: "must respond to %{name}"
 
       lt?: "must be less than %{num}"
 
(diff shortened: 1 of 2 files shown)


esparta:
Even though the message is more accurate, IMO it's a breaking change because the users may be expecting this previous specific error message on their own tests.

solnic:
actually, this predicate shouldn't exist anymore. I forgot to remove it prior 1.0.0 release. Geez 🤦

tylerhunt:
@solnic Hmm, how should this be handled instead? What was the deciding factor in choosing which predicates would be included?

solnic:
@tylerhunt `attr?` was added back in dry-validation 0.x era when we still thought that validating objects could be a use case that dry-validation should support. This has changed and now both dry-validation and dry-schema are meant to validation data only, that is, objects that represent data. The interface for accessing the values is `#[]`, checking method names, aka attributes, is not a feature we want to support because it's against the philosophy both libraries promote.

I'm sorry for this but I'll close this PR and make a note that this predicate and everything related should be removed in 2.0.0.

PR #337, opened 2020-12-27
Add specs using fallbacks

flash-gordon:
This is based on https://github.com/dry-rb/dry-types/pull/410 and https://github.com/dry-rb/dry-types/pull/411 It looks like the initial plan played out quite well but I'll be making more tests today (next step is dry-struct).


pyromaniac (commented):
I just love it!

PR #341, opened 2021-01-22
Use Messages#[] to translate unexpected keys

robhanlon22:
[changelog]

version: unreleased
fixed: "Messages#[] handles meta/no meta cases more gracefully and has better interoperability with the I18n backend. This brings MessageCompiler#visit_unexpected_key up to parity with MessageCompiler#visit_predicate. Uses visit_predicate as basis for visit_unexpected_key. (@robhanlon22)"


robhanlon22:
This is an empty PR to update the change log.

PR #348, opened 2021-04-08
Add error message to respond_to? predicate

rindek:
This was not reported on https://discourse.dry-rb.org/, though I think it can be treated as "typo", thus just straight PR:

When using Types.Interface, error is not produced in the `Dry::Schema::Result`

Ex.

```
Dry::Schema.Params { required(:a).filled(Types.Interface(:inv)) }.call(a: 1)
=> #<Dry::Schema::Result:0x1680>
```

The reason behind that is because errors message compiler can't find translation:
```
Dry::Schema::MissingMessageError (Message template for :a under "" was not found. Searched in:)
"en.dry_schema.errors.rules.a.respond_to?.arg.default"
"en.dry_schema.errors.rules.a.respond_to?"
"en.dry_schema.errors.respond_to?.failure"
"en.dry_schema.errors.respond_to?.value.a"
"en.dry_schema.errors.respond_to?.value.default.arg.default"
"en.dry_schema.errors.respond_to?.value.default"
"en.dry_schema.errors.respond_to?.arg.default"
"en.dry_schema.errors.respond_to?"
```

Expected:
```
Dry::Schema.Params { required(:a).filled(Types.Interface(:inv)) }.call(a: 1)
=> #<Dry::Schema::Result{:a=>1} errors={:a=>["must respond to inv"]}>
```

Adding a translation to `respond_to?` fixes the above issue

### DIFF
--- a/config/errors.yml
+++ b/config/errors.yml
@@ -87,6 +87,8 @@ en:
 
       type?: "must be %{type}"
 
+      respond_to?: "must respond to %{method}"
+
       size?:
         arg:
           default: "size must be %{size}"
--- a/lib/dry/schema/macros/dsl.rb
+++ b/lib/dry/schema/macros/dsl.rb
@@ -17,6 +17,7 @@ module Dry
 
         undef :eql?
         undef :nil?
+        undef :respond_to?
 
         # @!attribute [r] chain
         #   Indicate if the macro should append its rules to the provided trace
--- a/spec/integration/schema/predicates/respond_to_spec.rb
+++ b/spec/integration/schema/predicates/respond_to_spec.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 RSpec.describe 'Predicates: Respond To' do
-  xcontext 'with required' do
+  context 'with required' do
     subject(:schema) do
       Dry::Schema.define { required(:foo) { respond_to?(:bar) } }
     end
@@ -47,7 +47,7 @@ RSpec.describe 'Predicates: Respond To' do
     end
   end
 
-  xcontext 'with optional' do
+  context 'with optional' do
     subject(:schema) do
       Dry::Schema.define { optional(:foo) { respond_to?(:bar) } }
     end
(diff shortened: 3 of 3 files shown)


solnic (commented):
Thanks, this makes sense. Any chance you could create a simple spec under spec/integration/schema/predicates/respond_to_spec.rb?

rindek:
@solnic I've added tests but I managed to test only macro-way, as I found a bug

It looks like defining schema passing predicate directly to require/optional is broken

```
Dry::Schema.define { required(:foo) { respond_to?(:bar) } }
NoMethodError: undefined method `to_ast' for false:FalseClass
Did you mean?  to_s
from /Users/rindek/projects/dry-rb/dry-schema/lib/dry/schema/trace.rb:89:in `map'
```

My guess is that `respond_to?` in this case is calling `Object#respond_to?` instead of predicate definition from `dry-logic`

I've skipped failing tests, although I'm not sure what should be done with that bug

solnic:
> My guess is that respond_to? in this case is calling Object#respond_to? instead of predicate definition from dry-logic

Correct. If you add `undef :respond_to?` here "lib/dry/schema/macros/dsl.rb" then it'll work.

rindek:
@solnic I've added undef in the dsl, tests are green on MRI, but some fail on jruby (others than what I added)
Unfortunately I don't know jruby that much to figure out what exactly is the root cause, do you maybe have an idea?

solnic:
> some fail on jruby (others than what I added)
Unfortunately I don't know jruby that much to figure out what exactly is the root cause, do you maybe have an idea?

thanks, this is confusing - I need to debug this under jruby :( The truth is, we should replace `respond_to?` predicate with some non-conflicting name. It was reported at least once that this predicate caused issues. Definitely something to do in 2.0.0.

flash-gordon:
I'll cut a minor release in dry-logic, this should fix jruby failures https://github.com/dry-rb/dry-logic/pull/85

solnic (approved):

PR #353, opened 2021-05-07
Fixup nested step callbacks & non-hash input handling 

### LINKED ISSUE #350: Processor steps callbacks for nested schema are called in wrong order
## Describe the bug

```ruby
require 'dry-schema'
require 'json'

field_schema = Dry::Schema.define do
  required(:string_field)

  before(:key_coercer) { |result| result.to_h }
end

  required(:schema_field).filled(field_schema)

  before(:key_coercer) do |result|
    JSON.parse(result.output, {symbolize_names: true})
  end
end

p schema.call({ schema_field: { string_field: 'foo' } }.to_json)
```

## To Reproduce

Run example above, 
```
/home/ojab/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/dry-schema-1.6.2/lib/dry/schema/result.rb:52:in `[]': no implicit conversion of Symbol into Integer (TypeError)
	from /home/ojab/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/dry-schema-1.6.2/lib/dry/schema/result.rb:52:in `block in at'
	from /home/ojab/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/dry-schema-1.6.2/lib/dry/schema/path.rb:85:in `each'
	from /home/ojab/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/dry-schema-1.6.2/lib/dry/schema/path.rb:85:in `each'
	from /home/ojab/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/dry-schema-1.6.2/lib/dry/schema/result.rb:52:in `reduce'
…
```
We're not applying `schema.before(:key_coercer)` before trying to get `result[:schema_field]` for `field_schema`, which understandably fails since result contains unparsed string. 

## Expected behavior

It works! 

## My environment

- Affects my production application: NO
- Ruby version: 3.0.1
- OS: Linux x86_64

ojab:
Fixes #350 
Also some drive-by simplifications to `Path`

dry-v specs are passing with this changes.

### DIFF
--- a/lib/dry/schema/step.rb
+++ b/lib/dry/schema/step.rb
@@ -33,11 +33,23 @@ module Dry
 
       # @api private
       def call(result)
-        scoped_result = path.equal?(EMPTY_PATH) ? result : result.at(path)
+        scoped_results(result).each do |scoped_result|
+          output = executor.(scoped_result)
+          scoped_result.replace(output) if output.is_a?(Hash)
+        end
+      end
 
-        output = executor.(scoped_result)
-        scoped_result.replace(output) if output.is_a?(Hash)
-        output
+      # @api private
+      def scoped_results(initial_result)
+        path.reduce([initial_result]) do |results, path_key|
+          results.flat_map do |result|
+            scoped_result = result.at(path_key)
+
+            next scoped_result unless scoped_result.output.is_a?(Array)
+
+            Array.new(scoped_result.output.length) { |index| scoped_result.at(index) }
+          end
+        end
       end
 
       # @api private
--- a/lib/dry/schema/step.rb
+++ b/lib/dry/schema/step.rb
@@ -43,12 +43,14 @@ module Dry
       def scoped_results(initial_result)
         path.reduce([initial_result]) do |results, path_key|
           results.flat_map do |result|
+            next unless result.output.key?(path_key)
+
             scoped_result = result.at(path_key)
 
             next scoped_result unless scoped_result.output.is_a?(Array)
 
             Array.new(scoped_result.output.length) { |index| scoped_result.at(index) }
-          end
+          end.compact
         end
       end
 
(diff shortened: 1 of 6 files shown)


lib/dry/schema/step.rb
```diff
@@ -17,53 +17,30 @@ class Step
...
+          scoped_result.replace(output) if output.is_a?(Hash)
+          output
+        end
       end
 
       # @api private
-      def scoped(path)
-        Scoped.new(path, self)
+      def scoped(new_path)
+        path.keys.unshift(*new_path)
```
ojab, line 42: We're mutating step here, but right now it doesn't matter and easy change if would be needed later.

ojab: Ok, it actually breaks reused schemas, fixed.

ojab:
Added handling of schemas inside arrays. Will break this to multiple PRs/squash/whatever if needed.

I don't quite like how it works, but dunno if it could be done undoubtly better.
IMHO would be easier and more clear to convert key_coercer/filter_schema/value_coercer to underlying type's constructor modifiers instead.

ojab:
Found/fixed an issue with callbacks for missing `optional.schema`, also there is a noticeable hit in microbenchmarks after this PR, will take a look sometime this week.

solnic:
Hey man thanks for this PR, it's awesome that you started tackling this. FYI `Path` and related things have been on a list of things to refactor/clean-up/improve for a very long time now. I'm more than happy to merge your stuff but please create separate PRs with your work split into refactors and bug fixes. This lib is foundational so we need to be more careful with changes.

solnic:
@ojab so I guess this can be closed now, right? 🙂

ojab:
Yeah, remaining part is array handling, I'll prepare a clean new PR with that. Thanks!

solnic:
@ojab oh damn I thought everything was covered and I already released it 🙂

PR #356, opened 2021-05-24
Update for latest dry configurable

timriley:


solnic (commented):
The same case as with dry-validation - I think it'd be good to make it work with both versions of dry-configurable.

solnic:
so, the patch won't work because once you start nesting setting the execution context changes, gah

timriley:
This one turned out not to need any kind of special patching now that we made sure the upcoming dry-configurable deprecations covered all possible cases.

PR #358, opened 2021-06-08
Simplify `Dry::Schema::Path`

ojab:
First extract from https://github.com/dry-rb/dry-schema/pull/353.

master dry-v specs are passing locally, I don't see any changes in benchmarks.

### DIFF
--- a/lib/dry/schema/path.rb
+++ b/lib/dry/schema/path.rb
@@ -71,24 +71,13 @@ module Dry
       end
 
       # @api private
-      def index(key)
-        keys.index(key)
-      end
-
-      # @api private
       def include?(other)
         keys[0, other.keys.length].eql?(other.keys)
       end
 
       # @api private
       def <=>(other)
-        raise ArgumentError, "Can't compare paths from different branches" unless same_root?(other)
-
-        return 0 if keys.eql?(other.keys)
-
-        res = key_matches(other).compact.reject { |value| value.equal?(false) }
-
-        res.size < count ? 1 : -1
+        keys.length <=> other.keys.length if include?(other) || other.include?(self)
       end
 
       # @api private
@@ -96,11 +85,6 @@ module Dry
         self.class.new(
           keys.take_while.with_index { |key, index| other.keys[index].eql?(key) }
         )
-      end
-
-      # @api private
-      def key_matches(other, meth = :map)
-        public_send(meth) { |key| (idx = other.index(key)) && keys[idx].equal?(key) }
       end
 
       # @api private
--- a/lib/dry/schema/path.rb
+++ b/lib/dry/schema/path.rb
@@ -93,12 +93,8 @@ module Dry
 
       # @api private
       def &(other)
-        unless same_root?(other)
-          raise ArgumentError, "#{other.inspect} doesn't have the same root #{inspect}"
-        end
-
         self.class.new(
-          key_matches(other, :select).compact.reject { |value| value.equal?(false) }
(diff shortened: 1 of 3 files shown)


ojab:
2.5 specs fail to run due to http://github.com/dry-rb/dry-schema/commit/bf3757f0a2764b1b821c564b30a205e138d11417

lib/dry/schema/path.rb
```diff
@@ -60,84 +60,33 @@ def initialize(keys)
...
       def each(&block)
         keys.each(&block)
       end
 
-      # @api private
-      def index(key)
-        keys.index(key)
-      end
-
-      def without_index
```
ojab, line 93: It's not marked as `@private`, but it's unused in dry-s/dry-v and I'm not sure if it's intended to be public.

solnic:
@ojab please rebase, I updated CI config in master

ojab:
Done, CI is green.

lib/dry/schema/path.rb
```diff
@@ -60,84 +60,33 @@ def initialize(keys)
...
-              else
-                EMPTY_HASH.dup
-              end
-
-          curr_idx += 1
-        end
+        value = [value] unless value.is_a?(Array)
 
-        hash
+        keys.reverse_each.reduce(value) { |result, key| {key => result} }
```
solnic, line 65: Any chance you could compare this wonderful one-liner with the horrible `while` loop? I'm asking because there are places like this where these procedural loops are used purely due to performance reasons. I don't remember exactly but I'm quite certain that in this case the difference was huge. This hash is produced when generating messages and it's literally the biggest performance bottleneck in dry-schema (thus in dry-validation too).

ojab:
```ruby
require "benchmark/ips"
require 'dry-schema'


path = Dry::Schema::Path.new(Array.new(1000) { ['a'..'z'].sample })

Benchmark.ips do |x|
  x.report('old impl') { path.old_to_h }
  x.report('new impl') { path.to_h }

  x.compare!
end
```

```
            old impl    70.000  i/100ms
            new impl    85.000  i/100ms
            old impl    764.275  (± 2.1%) i/s -      3.850k in   5.039660s
            new impl    875.548  (± 2.3%) i/s -      4.420k in   5.050804s

Comparison:
            new impl:      875.5 i/s
            old impl:      764.3 i/s - 1.15x  (± 0.00) slower
```
/shrug (3.0.1 & 2.7.2 looks similar)

I tried to find out where exactly `Path` operations is a hot path during actual schema validation or error/to_h generation, but failed to do so, thus I doubt that rather minor performance changes are meaningful in the real life anyway.

ojab: In regard to message generation:
```ruby
require 'benchmark/ips'
require 'dry-schema'

PersonSchema = Dry::Schema.Params do
  required(:foo).hash do
    required(:bar).hash do
      required(:name).value(:string)
      required(:age).value(:integer, gteq?: 18)
      required(:email).value(:string)
    end
  end
end

INVALID_INPUT = {
  foo: {
    bar: {
      name: :John,
      age: "17",
      email: nil
    }.freeze
  }.freeze
}.freeze

1_000_000.times { PersonSchema.(INVALID_INPUT).errors.to_h }
```
Says something like this to me
flamegraph

ojab: When checking `result.error?(foo: {bar: :name})` in the loop:
flamegraph

solnic:
Thanks, this is really great. I hope this doesn't break anything but in the test suite we trust.

PR #360, opened 2021-06-10
Merge `Dry::Schema::Step::Scoped` into `Dry::Schema::Step`

ojab:
Second extract from https://github.com/dry-rb/dry-schema/pull/353

It makes things more streamlined. And preparational commit:
```
   Drop `Dry::Schema::Result#results`
    
    It's unused, we're working only with result_ast and if value coercer
    is not run on the Result (will be the case in the next commit) — it's
    frozen after `.new`, so `@result_ast ||=` fails with `FrozenError` if
    we try to get `#errors`, for example
    
    Also fix the typo `param` -> `option`
```

Unfortunately can't invent not-noisy benchmark, so tested by this:
```ruby
VALID_INPUT = {
  name: "John",
  age: 20,
  email: "john@doe.com"
}.freeze

INVALID_INPUT = {
  name: :John,
  age: "17",
  email: nil
}.freeze

MESSAGE_COMPILER = PersonSchema.(VALID_INPUT.dup).message_compiler.freeze

def result(input)
  Dry::Schema::Result.method(:new).super_method.call(input.dup, message_compiler: MESSAGE_COMPILER)
end

key_map = Dry::Schema::KeyMap.new([[:name, :age, :email]])
coercer =  Dry::Schema::KeyCoercer.new(key_map, &:to_sym)
step = Dry::Schema::Step.new(executor: coercer, type: :core, name: :key_coercer)

Benchmark.ips do |x|
  x.config(stats: :bootstrap, confidence: 95)

  x.report("valid input") { step.(result(VALID_INPUT)) }
  x.report("invalid input") { step.(result(INVALID_INPUT)) }
  x.compare!
end
```
Before:
```
         valid input    11.959k i/100ms
       invalid input    12.077k i/100ms
         valid input    113.668k (± 1.1%) i/s -    574.032k in   5.059266s
       invalid input    122.840k (± 1.2%) i/s -    615.927k in   5.027367s
                   with 95.0% confidence

Comparison:
       invalid input:   122840.2 i/s
         valid input:   113668.2 i/s - 1.08x  (± 0.02) slower
                   with 95.0% confidence
```
After:
```
         valid input    11.004k i/100ms
       invalid input    12.469k i/100ms
         valid input    117.272k (± 1.2%) i/s -    594.216k in   5.077576s
       invalid input    114.757k (± 1.1%) i/s -    573.574k in   5.007477s
                   with 95.0% confidence

Comparison:
         valid input:   117272.0 i/s
       invalid input:   114757.4 i/s - same-ish: difference falls within error
                   with 95.0% confidence
```

In theory both valid & invalid inputs should be the same, since we're not generating error messages, but  there is variation. The same with before/after comparison, results are comparable, but different runs show `115000 i/s ± 10%`.

dry-v specs are passing.

### DIFF
--- a/lib/dry/schema/processor.rb
+++ b/lib/dry/schema/processor.rb
@@ -84,14 +84,14 @@ module Dry
       #
       # @api public
       def call(input)
-        Result.new(input, input: input, message_compiler: message_compiler) do |result|
+        Result.new(input.dup, message_compiler: message_compiler) do |result|
           steps.call(result)
         end
       end
       alias_method :[], :call
 
       # @api public
-      def xor(other)
+      def xor(_other)
         raise NotImplementedError, "composing schemas using `xor` operator is not supported yet"
       end
       alias_method :^, :xor
--- a/lib/dry/schema/step.rb
+++ b/lib/dry/schema/step.rb
@@ -7,6 +7,9 @@ module Dry
   module Schema
     # @api private
     class Step
+      EMPTY_PATH = Path.new([]).freeze
+      private_constant :EMPTY_PATH
+
       # @api private
       attr_reader :name
 
@@ -17,53 +20,34 @@ module Dry
       attr_reader :executor
 
       # @api private
-      class Scoped
-        # @api private
-        attr_reader :path
-
-        # @api private
-        attr_reader :step
-
-        # @api private
-        def initialize(path, step)
-          @path = Path[path]
-          @step = step
-        end
-
-        # @api private
-        def scoped(new_path)
-          self.class.new(Path[[*new_path, *path]], step)
-        end
-
-        # @api private
-        def call(result)
-          result.at(path) do |scoped_result|
-            output = step.(scoped_result).to_h
-            target = Array(path)[0..-2].reduce(result) { |a, e| a[e] }
-
-            target.update(path.last => output)
-          end
-        end
-      end
+      attr_reader :path
 
       # @api private
-      def initialize(type:, name:, executor:)
+      def initialize(type:, name:, executor:, path: EMPTY_PATH)
         @type = type
         @name = name
         @executor = executor
+        @path = path
         validate_name(name)
       end
 
       # @api private
       def call(result)
-        output = executor.(result)
-        result.replace(output) if output.is_a?(Hash)
+        scoped_result = path.equal?(EMPTY_PATH) ? result : result.at(path)
+
+        output = executor.(scoped_result)
+        scoped_result.replace(output) if output.is_a?(Hash)
         output
       end
 
       # @api private
-      def scoped(path)
-        Scoped.new(path, self)
+      def scoped(parent_path)
+        self.class.new(
+          type: type,
+          name: name,
+          executor: executor,
+          path: Path.new([*parent_path, *path])
+        )
       end
 
       private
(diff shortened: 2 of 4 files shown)


lib/dry/schema/processor.rb
```diff
@@ -84,7 +84,7 @@ def new(options = nil, &block)
       #
       # @api public
       def call(input)
-        Result.new(input, input: input, message_compiler: message_compiler) do |result|
+        Result.new(input.dup, message_compiler: message_compiler) do |result|
```
ojab, line 87: Previously we were mutating only inner hashes of input, now we're using `Hash#replace` as a `Dry::Schema::Result#replace`, so non-frozen input is needed.

lib/dry/schema/step.rb
```diff
@@ -17,53 +20,34 @@ class Step
...
-
-            target.update(path.last => output)
-          end
-        end
-      end
+      attr_reader :path
 
       # @api private
-      def initialize(type:, name:, executor:)
+      def initialize(type:, name:, executor:, path: EMPTY_PATH)
```
ojab, line 26: No `#dup` here since we want to use blazing fast `#equal?` below instead of slow `#eql?` or whatever else.
We're not mutating path anyway and it's frozen, so :thisisfine:.

Knowing about implementation details we should froze `EMPTY_PATH.keys`, but let's pretend that we don't know it, if someone wanna mess up with internals of our private constant — all bets are off.

lib/dry/schema/step.rb
```diff
@@ -17,53 +20,34 @@ class Step
...
 
       # @api private
-      def scoped(path)
-        Scoped.new(path, self)
+      def scoped(parent_path)
+        self.class.new(
+          type: type,
+          name: name,
+          executor: executor,
+          path: Path.new([*parent_path, *path])
```
ojab, line 49: This is no different from old `Step::Scoped#scoped` performance-wise.

lib/dry/schema/step.rb
```diff
@@ -17,53 +20,34 @@ class Step
...
         @executor = executor
+        @path = path
         validate_name(name)
       end
 
       # @api private
       def call(result)
-        output = executor.(result)
-        result.replace(output) if output.is_a?(Hash)
+        scoped_result = path.equal?(EMPTY_PATH) ? result : result.at(path)
```
ojab, line 36: `result.at(path) do |scoped_result|` would be better, but way slower for the general/empty_path usecase.

I think that I could alleviate the hit in the latter extracted PRs, but not sure yet.

ojab:
BTW, is there any easy way to run rubocop locally like it run on CI?

lib/dry/schema/step.rb
```diff
@@ -17,53 +20,34 @@ class Step
...
+
+        output = executor.(scoped_result)
+        scoped_result.replace(output) if output.is_a?(Hash)
```
ojab, line 39: Previously this check was missing from `Scoped#call`, so we have a bit different behavior now. I don't think it's an issue.

solnic (approved):

solnic:
Thanks, I appreciate that you explained things in the comments. This is a bold refactoring but again, in the test suite we trust.

PR #361, opened 2021-06-11
Run outer schema processor steps before inner ones

ojab:
Extract from https://github.com/dry-rb/dry-schema/pull/353.

Pretty straightforward, just sorting steps by shortest path first on schema creation. 
We need to compare paths with different roots for this, so allow it.

### DIFF
--- a/lib/dry/schema/processor_steps.rb
+++ b/lib/dry/schema/processor_steps.rb
@@ -87,6 +87,7 @@ module Dry
       def after(name, &block)
         after_steps[name] ||= EMPTY_ARRAY.dup
         after_steps[name] << Step.new(type: :after, name: name, executor: block)
+        after_steps[name].sort_by!(&:path)
         self
       end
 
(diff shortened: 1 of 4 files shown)


lib/dry/schema/processor_steps.rb
```diff
@@ -120,18 +122,20 @@ def merge(other)
...
+          before_steps[name].concat(steps.map { |step| step.scoped(path) }).sort_by!(&:path)
         end
 
         other.after_steps.each do |name, steps|
-          (after_steps[name] ||= []).concat(steps.map { |step| step.scoped(path) })
+          after_steps[name] ||= []
+          after_steps[name].concat(steps.map { |step| step.scoped(path) }).sort_by!(&:path)
         end
       end
     end
```
ojab, line 141: Probably we could sort callbacks just once instead of on each import/merge, but we're doing it only on schema creation, thus probably doesn't warrant additional complexity.

Issue #355, opened 2021-05-20
validate_keys=true ignored when the value is array


dgutov:
## Describe the bug

When the unexpected key's value is an array, it is not reported as unexpected. Validation can be marked as successful even.

## To Reproduce

```rb
$ irb
irb(main):001:0> require 'dry-schema'
=> true
irb(main):002:0> require 'dry/schema/version'
=> true
irb(main):003:0> Dry::Schema::VERSION
=> "1.6.2"
irb(main):004:1* schema = Dry::Schema.define do
irb(main):005:1*   config.validate_keys = true
irb(main):006:1*   
irb(main):007:2*   optional(:address).hash do
irb(main):008:1*   end
irb(main):009:0> end
=> #<Dry::Schema::Processor keys=[{:address=>[]}] rules={:address=>"key?(:address) THEN key[addr...
irb(main):010:0> schema.call(address: {a: 1})
=> #<Dry::Schema::Result{:address=>{}} errors={:address=>{:a=>["is not allowed"]}}>
irb(main):011:0> schema.call(address: {a: [1, 2, 3]})
=> #<Dry::Schema::Result{:address=>{}} errors={}>
irb(main):012:0> schema.call(address: {a: {b: 1}})
=> #<Dry::Schema::Result{:address=>{}} errors={:address=>{:a=>{:b=>["is not allowed"]}}}>
irb(main):013:0> schema.call(address: {a: 'zz'})
=> #<Dry::Schema::Result{:address=>{}} errors={:address=>{:a=>["is not allowed"]}}>
```

## Expected behavior

It should return `#<Dry::Schema::Result{:address=>{}} errors={:address=>{:a=>["is not allowed"]}}>` in all cases.

BTW, the nested hash behavior also looks wrong. It should not report that the `:b` key is out of place, it should do that for the parent structure instead. IOW, I expect all of the cases above to return the same value.

## My environment

- Affects my production application: **YES**
- Ruby version: 2.7.2
- OS: Ubuntu 20.04

Issue #349, opened 2021-04-13
Empty hashes not detected by validate_keys is true


patodevilla:
## Describe the bug

It appears empty hashes are not detected by `validate_keys`

When setting `validate_keys` as true, empty hashes are being ignored by the validator

## To Reproduce

```ruby
schema = Dry::Schema.JSON do
  config.validate_keys = true
end
schema.({hello: {}})
 => #<Dry::Schema::Result{} errors={}> 
schema.({hello: 1})
 => #<Dry::Schema::Result{} errors={:hello=>["is not allowed"]}>
```

## Expected behavior

Unexpected key `hello` should be in the errors hash regardless of the key's value

## My environment

- Affects my production application: **YES**
- Ruby version: ruby-2.5.7

PR #363, opened 2021-06-17
Fix key validator false negatives on empty collections

Drenmi:
### What was happening?

As described in #349 and #355, when using `validate_keys = true`, there are false negatives when the schema encounters unexpected keys where the value is an empty array or -hash, or an array with only primitives.

This was happening because the recursively defined `KeyValidator#key_paths` did not handle the cases where there was nothing to recurse over. (The correct behaviour would be to instantly return the key.)

The output of this method is what is subsequently checked against the key map, but with the relevant keys missing, the validation logic would overlook these unexpected keys.

### How does this fix it?

- When encountering an empty hash value, immediately add the key to the key paths.
- When encountering an array value, return the key immediately if the array 1) has no nested collections, 2) all nested collections are empty, or 3) contains only primitives.

### What is the impact?

This change affects only private APIs, but the obvious "side effect" is that the bug is fixed, and anyone whose application incidentally works because unexpected keys were ignored, might start seeing errors in their application after upgrading.

### Limitations

There was a related, but separate, issue highlighted in #355. For nested hashes, the error is tacked onto the innermost hash, rather than on the first key that is unexpected. After having a quick look I decided to not address that in this same PR, as the changes are quite involved, and probably deserve their own PR.

### DIFF
--- a/lib/dry/schema/key_validator.rb
+++ b/lib/dry/schema/key_validator.rb
@@ -51,9 +51,14 @@ module Dry
         hash.flat_map { |key, _|
           case (value = hash[key])
           when Hash
+            next key.to_s if value.empty?
+
             [key].product(key_paths(hash[key])).map { |keys| keys.join(DOT) }
           when Array
-            hashes_or_arrays = value.select { |e| e.is_a?(Array) || e.is_a?(Hash) }
+            hashes_or_arrays = value.select { |e| (e.is_a?(Array) || e.is_a?(Hash)) && !e.empty? }
+
+            next key.to_s if hashes_or_arrays.empty?
+
             hashes_or_arrays.flat_map.with_index { |el, idx|
               key_paths(el).map { |path| ["#{key}[#{idx}]", *path].join(DOT) }
             }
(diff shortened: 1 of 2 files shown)


lib/dry/schema/key_validator.rb
```diff
@@ -51,9 +51,14 @@ def key_paths(hash)
...
           when Hash
+            next key.to_s if value.empty?
+
             [key].product(key_paths(hash[key])).map { |keys| keys.join(DOT) }
           when Array
-            hashes_or_arrays = value.select { |e| e.is_a?(Array) || e.is_a?(Hash) }
+            hashes_or_arrays = value.select { |e| (e.is_a?(Array) || e.is_a?(Hash)) && !e.empty? }
+
+            next key.to_s if hashes_or_arrays.empty?
+
```
solnic, line 61: since it's skipping empy items now, aren't we going to end up with `nil`s in the resulting array?

Drenmi: The `next` keyword takes one argument which will be taken by the iterator. In this case that would be `key.to_s`. 🙂 

```ruby
3.times.map { |n| next n }
#=> [0, 1, 2]
```

solnic: Ah doooh of course! Thanks for the explanation 😄

solnic (approved):

PR #364, opened 2021-06-27
Prevent error message YAML files from being parsed multiple times

### LINKED ISSUE #352: Each schema hold it's own copy of all the validation strings
## Describe the bug

Each schema instance load it's own version of the validations strings. It's both a waste of CPU during the application boot as the file has to be parsed repeatedly, and a waste of RAM.

Here's a heap-profiler report:

```
   2.48 kB      62  "must be odd"
                62  msgpack-1.4.2/lib/msgpack/factory.rb:70
 
   2.48 kB      62  "must be filled"
                61  msgpack-1.4.2/lib/msgpack/factory.rb:70
                 1  bootsnap-1.7.5/lib/bootsnap/compile_cache/iseq.rb:19
 
   2.48 kB      62  "must be even"
                62  msgpack-1.4.2/lib/msgpack/factory.rb:70
 
   2.48 kB      62  "must be boolean"
                61  msgpack-1.4.2/lib/msgpack/factory.rb:70
                 1  bootsnap-1.7.5/lib/bootsnap/compile_cache/iseq.rb:19
 
   2.48 kB      62  "must be an integer"
                61  msgpack-1.4.2/lib/msgpack/factory.rb:70
                 1  bootsnap-1.7.5/lib/bootsnap/compile_cache/iseq.rb:19
 
   2.48 kB      62  "must be an array"
                61  msgpack-1.4.2/lib/msgpack/factory.rb:70
                 1  bootsnap-1.7.5/lib/bootsnap/compile_cache/iseq.rb:19
 
   2.48 kB      62  "must be a string"
                61  msgpack-1.4.2/lib/msgpack/factory.rb:70
                 1  bootsnap-1.7.5/lib/bootsnap/compile_cache/iseq.rb:19
 
   2.48 kB      62  "must be a float"
                61  msgpack-1.4.2/lib/msgpack/factory.rb:70
                 1  bootsnap-1.7.5/lib/bootsnap/compile_cache/iseq.rb:19
```

These strings are reported as created by messagepack because of Bootsnap's YAML precompilation cache, but I tracked them down to `dry-schema`. From this extract you can likely guess that we have 61 schemas.

## To Reproduce

Simply creating multiple schema in the same process should exhibit the behavior.

## Expected behavior

I would expect schemas to somehow have a mechanism to re-use already loaded error messages.

Or at the very least, `dry-schema` could pass `freeze: true` to `YAML.load_file` so that the strings would be deduplicated.

alassek:
Fixes #352

Adds `source_cache` to `Dry::Schema::Messages::YAML`.

Since this data is now being shared by Schemas, I opted to freeze it.

I'm leaning towards not adding test coverage for this specifically, since it's changing the implementation of a feature that is already under test.

### DIFF
--- a/lib/dry/schema/messages/yaml.rb
+++ b/lib/dry/schema/messages/yaml.rb
@@ -72,6 +72,11 @@ module Dry
       end
 
       # @api private
+      def self.source_cache
+        @source_cache ||= Concurrent::Map.new
+      end
+
+      # @api private
       def initialize(data: EMPTY_HASH, config: nil)
         super()
         @data = data
@@ -179,7 +184,9 @@ module Dry
 
       # @api private
       def load_translations(path)
-        data = self.class.flat_hash(YAML.load_file(path))
+        data = self.class.source_cache.fetch_or_store(path) do
+          self.class.flat_hash(YAML.load_file(path)).freeze
+        end
 
         return data unless custom_top_namespace?(path)
 


solnic (approved):
Thanks!

Issue #342, opened 2021-01-28
Composing Schemas Raises NoMethodError in Validation Failures


nbibler:
## Describe the bug

[Composing complex Schemas](https://dry-rb.org/gems/dry-schema/1.5/advanced/composing-schemas/) appears to throw a `NoMethodError: undefined method 'to_or' for #<Array:0x0>` in a failure case rather than simply return an object with populated validation errors/failures.

This was found while using:

```
Using bundler 2.1.4
Using dry-equalizer 0.3.0
Using dry-initializer 3.0.4
Using concurrent-ruby 1.1.8
Using dry-inflector 0.2.0
Using ice_nine 0.11.2
Using dry-core 0.5.0
Using dry-configurable 0.12.0
Using dry-logic 1.1.0
Using dry-container 0.7.2
Using dry-types 1.5.0
Using dry-schema 1.6.0
Using dry-struct 1.4.0
Using dry-validation 1.6.0
```

## To Reproduce

```ruby
#!/usr/bin/env ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'
  gem 'dry-equalizer'
  gem 'dry-schema'
  gem 'dry-struct'
  gem 'dry-types'
  gem 'dry-validation'
end

MatchingQuestion = Dry::Schema.Params do
  required(:attributes).hash do
    required(:value).value(:array, :filled?).each(:str?, format?: /\A\d+\Z/)
  end
  required(:relationships).hash do
    required(:question).hash do
      required(:data).hash do
        required(:type).value(:str?, eql?: 'matching')
      end
    end
  end
end

MultipleChoiceQuestion = Dry::Schema.Params do
  required(:attributes).hash do
    required(:value).value(:array, :filled?).each(:str?, format?: /\A\d+\Z/)
  end
  required(:relationships).hash do
    required(:question).hash do
      required(:data).hash do
        required(:type).value(:str?, eql?: 'multiple-choice')
      end
    end
  end
end

Question = Dry::Schema.Params do
  required(:data).hash(MatchingQuestion | MultipleChoiceQuestion)
end

def validate(data)
  p Question.(data).errors.to_h
rescue
  p $!.inspect
end

# Good data, no errors/exceptions
validate(data: {attributes: {value: ['1', '2']}, relationships: {question: {data: {type: 'matching'}}}})
[...]

PR #366, opened 2021-06-29
Fix disjunction compose with schemas nesting

vsuhachev:
Try to fix this issue #342


lib/dry/schema/message/or/multi_path.rb
```diff
@@ -20,14 +20,14 @@ class MultiPath < Abstract
           def initialize(*args)
             super
             @root = [left, right].flatten.map(&:_path).reduce(:&)
-            @left = left.map { |msg| msg.to_or(root) }
-            @right = right.map { |msg| msg.to_or(root) }
+            @left = left.flatten.map { |msg| msg.to_or(root) }
+            @right = right.flatten.map { |msg| msg.to_or(root) }
```
solnic, line 24: Wouldn't `flat_map` be enough here? `Array#flatten` with no arguments does deep flattening.

vsuhachev: Each schema nesting level comes with same arrays nesting. I tested it. Without deep flatten it not worked with same error

ianks (approved):

PR #371, opened 2021-08-29
Use explicit conversion of config to hash

timriley:
The support for implicit conversion has been removed (https://github.com/dry-rb/dry-configurable/pull/114) and will be included in the upcoming dry-configurable 0.13.0 release.

This change is also compatible with all existing releases of dry-configurable, since they all feature a `Config#to_h` method.

### DIFF
--- a/lib/dry/schema/processor.rb
+++ b/lib/dry/schema/processor.rb
@@ -53,7 +53,7 @@ module Dry
         # @api public
         def define(&block)
           @definition ||= DSL.new(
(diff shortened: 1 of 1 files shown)


timriley:
I'm going to merge to merge this in right away, since it is uncontroversial given the feedback in https://github.com/dry-rb/dry-configurable/pull/114, and then cut a 1.7.1 release from it. /cc @flash-gordon @solnic

PR #386, opened 2021-11-26
Make path matching in key validator more strict

### LINKED ISSUE #385: Keys validation ignores substrings of defined keys
## Describe the bug

In schemas with `validate_keys` enabled, any input key that is a substring of either a defined key (including nested) or a full key path in dot notation will pass the validation.

## To Reproduce

```ruby
  config.validate_keys = true

  optional(:name).filled(:string)
  optional(:address).hash do
    optional(:city).filled(:string)
  end
end

schema.call(am: 'Jane', city: 'NYC', dress: { cit: '?' })
=> #<Dry::Schema::Result{} errors={} path=[]>
```

## Expected behavior

```ruby
schema.call(am: 'Jane', city: 'NYC', dress: { cit: '?' })
=> #<Dry::Schema::Result{} errors={:am=>["is not allowed"], :city=>["is not allowed"], :dress=>{:cit=>["is not allowed"]}} path=[]>
```

## My environment

- Affects my production application: YES
- Ruby version: 2.7.5p203
- OS: Debian 10.11

pharmaceutics:
Fixes #385

Loose matching was added to allow partial key paths to pass the `validate_keys` validation when used with `maybe` hashes  (#309, #312). This change maintains that behaviour, but prevents paths with partial keys from passing.

Originally, went with a regex solution but it turned out to be pretty sluggish when benchmarked. This version doesn't seem to have any noticeable impact on performance, at least for simple cases I tested.

### DIFF
--- a/lib/dry/schema/key_validator.rb
+++ b/lib/dry/schema/key_validator.rb
@@ -41,13 +41,19 @@ module Dry
         if path[INDEX_REGEX]
           key = path.gsub(INDEX_REGEX, BRACKETS)
 
-          if key_paths.none? { _1.include?(key) }
+          if key_paths.none? { paths_match?(key, _1) }
             arr = path.gsub(INDEX_REGEX) { ".#{_1[1]}" }
             arr.split(DOT).map { DIGIT_REGEX.match?(_1) ? Integer(_1, 10) : _1.to_sym }
           end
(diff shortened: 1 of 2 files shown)


flash-gordon:
LGTM 👍

Issue #390, opened 2021-12-19
"NoMethodError" error raised when validating nil value with `maybe(:array)`


hieuk09:
## Describe the bug

When using `maybe(:array)`, if the value is `nil`, error is raised instead of returning success object.

## To Reproduce

```ruby
schema = Dry::Schema.JSON do
  required(:list).maybe(:array).each(:string)
end

schema.call(list: nil) # raise NoMethodError
```

## Expected behavior

It should be consistent with `maybe` behavior for other types and return a result object, something like:

```ruby
#<Dry::Schema::Result{:list=>nil} errors={} path=[]>
```

## My environment

- Affects my production application: **NO**
- Ruby version: 3.0.2
- OS: MacOS

flash-gordon:
I think it's just `maybe` doesn't compose with `each`, use the lesser-known DSL instead `required(:list).maybe(array[:string])`.

flash-gordon:
Alternatively:
```ruby
schema = Dry::Schema.JSON do
  required(:list).maybe(:array) do
    nil? | each(:string)
  end
end
```

hieuk09:
Thanks for the fast response! These options work for me. Do you mind if I update [the doc](https://github.com/dry-rb/dry-schema/blob/master/docsite/source/basics/macros.html.md#maybe) to include these so that people are less confused when using `maybe`?

flash-gordon:
@hieuk09 sure, feel free!

flash-gordon:
I don't think it's a bug in the first place. OTOH, it's safe to be changed, it seems.

PR #393, opened 2021-12-21
Add documentation on how maybe can be used with array

hieuk09:
So that people won't mistakenly use `maybe` wrongly like in https://github.com/dry-rb/dry-schema/issues/390


bautrey37:
There is inconsistent behaviour now, because on dry-validation it is possible to use a `rule.each` on a maybe value. https://github.com/dry-rb/dry-validation/pull/709

Maybe this should be fixed instead of documented a caveat instead.

PR #395, opened 2022-01-28
Add spec reproducing issue with custom type with and nested hashes validation

edgenard:
This is meant to reproduce an issue where custom types in nested hashes do not take effect. 
This is my first time with this project, I hope the spec is in the right place. If not I'd be happy to move it to the appropriate place.

This PR is related to this Issue: https://github.com/dry-rb/dry-schema/issues/396

I tried but could not figure out how to get it to pass.

### DIFF
--- a/spec/integration/schema/macros/hash_spec.rb
+++ b/spec/integration/schema/macros/hash_spec.rb
@@ -103,7 +103,7 @@ RSpec.describe "Macros #hash" do
       subject(:schema) do
         ExpirationDate = Types::DateTime.constructor { |value| value.to_time.round.to_datetime }
         Dry::Schema.Params do
-          required(:unnested_dated).value(ExpirationDate)
+          required(:unnested_date).value(ExpirationDate)
           required(:foo).hash do
             required(:bar).hash do
               required(:nested_date).value(ExpirationDate)
@@ -112,7 +112,13 @@ RSpec.describe "Macros #hash" do
         end
       end
 
-      let(:input) { {foo: {nested_date: '2021-11-11T00:00:00+00:00'}, unnested_date: '2021-11-11T00:00:00+00:00'  } }
+      let(:input) do
+        {
+          foo: { bar: {nested_date: '2021-11-11T00:00:00+00:00'}},
+          unnested_date: '2021-11-11T00:00:00+00:00',
+        }
+        
+      end
 
       specify do
         expect(result).to be_successful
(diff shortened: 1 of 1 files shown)


spec/integration/schema/macros/hash_spec.rb
```diff
@@ -98,6 +98,28 @@
...
+          required(:unnested_dated).value(ExpirationDate)
+          required(:foo).hash do
+            required(:bar).hash do
+              required(:nested_date).value(ExpirationDate)
+            end
+          end
+        end
+      end
+
+      let(:input) { {foo: {nested_date: '2021-11-11T00:00:00+00:00'}, unnested_date: '2021-11-11T00:00:00+00:00'  } }
```
flash-gordon, line 115: it should be `foo: { bar:` according to the schema

flash-gordon (changes requested):
Thanks, I left a comment, can you also mark the spec as pending so that we can merge it in?

spec/integration/schema/macros/hash_spec.rb
edgenard: Thanks!

edgenard:
@flash-gordon  I tried to get the spec to pass but had a very hard time following the code, if you could point me in the right direction I'd be happy to try again. 

For example, what file or files would need to change to make it pass.

flash-gordon:
I didn't mean making it pass :) Just make it pending by changing specify to xspecify, it'll stop failing the CI.

edgenard:
> I didn't mean making it pass :) Just make it pending by changing specify to xspecify, it'll stop failing the CI.

I added the `pending` to the spec.

But I'd be happy to work on it if someone could point me in the right direction.

solnic:
Your type crashes, that's why it doesn't work:

```ruby
(byebug) Test::ExpirationDate["2021-11-11T00:00:00+00:00"]
*** Dry::Types::CoercionError Exception: undefined method `to_time' for "2021-11-11T00:00:00+00:00":String
Did you mean?  to_i
```

If you change it so that it does `DateTime.parse(value)`, the spec will pass.

@flash-gordon it seems like `Safe` constructor rescues silently this crash though - THAT is probably a bug 🙂

edgenard:
> Your type crashes, that's why it doesn't work:
> 
> ```ruby
> (byebug) Test::ExpirationDate["2021-11-11T00:00:00+00:00"]
> *** Dry::Types::CoercionError Exception: undefined method `to_time' for "2021-11-11T00:00:00+00:00":String
> Did you mean?  to_i
> ```
> 
> If you change it so that it does `DateTime.parse(value)`, the spec will pass.
> 
> @flash-gordon it seems like `Safe` constructor rescues silently this crash though - THAT is probably a bug 🙂

Thank you! I think that definitely cleared it up. And yeah the silent failure is certainly confusing to users of the library.

I'll close this PR.

PR #399, opened 2022-02-17
Deduplicate locale tokens

casperisfine:
```
   5.02 kB      57  "%<locale>s.dry_schema.rules.%<name>s"
                57  dry-schema-1.8.0/lib/dry/schema/messages/yaml.rb:43

   5.02 kB      57  "%<locale>s.dry_schema.errors"
                57  dry-schema-1.8.0/lib/dry/schema/messages/yaml.rb:41
```

### DIFF
--- a/lib/dry/schema/messages/yaml.rb
+++ b/lib/dry/schema/messages/yaml.rb
@@ -39,10 +39,10 @@ module Dry
           super do |config|
             config.default_locale = :en unless config.default_locale
 
-            config.root = "%<locale>s.#{config.root}"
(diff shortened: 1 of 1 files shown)


casperisfine:
[`String#-@`](https://ruby-doc.org/core-3.1.0/String.html#method-i-2D-40) interns the string.

As shown heap-profiler, if you instantiate X schemas, you end up with X copies of these two strings.

By using `String#-@` we deduplicate these strings so that all schemas point to the same (frozen) one, saving about 10KiB (in this case).

PR #404, opened 2022-04-14
fix: correct spelling error 'mininum' to 'minimum' in json-schema extension

svenanderzen:
## Describe the bug

👋

Seems like a small typo might have creep:de into the `json_schema` extension when using the `gteq?` predicate as it currently returns "mini_n_um" instead of "mini_m_um" 😄 

## To Reproduce

```ruby

UserSchema = Dry::Schema.JSON do
   required(:age).value(:integer, gteq?: 18)
end

require "json"

puts JSON.pretty_generate(UserSchema.json_schema)
```

Currently prints the following schema:

```json
{
  "type": "object",
  "properties": {
    "age": {
      "type": "integer",
      "mininum": 18
    }
  },
  "required": [
    "age"
  ]
}
```

## Expected behavior

To print the following schema:

```json
{
  "type": "object",
  "properties": {
    "age": {
      "type": "integer",
      "minimum": 18
    }
  },
  "required": [
    "age"
  ]
}
```

## Notes

Made a liberal interpretation of your contributing guidelines and assumed this qualifies as a "typo" and just went ahead and opened a pull request directly instead of creating an issue. Let me know if I should have done it the other way around 🙂 

Also, guess this could be something which affects downstream JSON schema validators (which uses the exported schema) since at least some of the validators will silently ignore the invalid "mini_n_um" constraint ([example](https://www.jsonschemavalidator.net/s/5YCuxssD)). Just something to be aware of I guess 🙂

### DIFF
--- a/lib/dry/schema/extensions/json_schema/schema_compiler.rb
+++ b/lib/dry/schema/extensions/json_schema/schema_compiler.rb
@@ -47,7 +47,7 @@ module Dry
             pattern: "^[0-9A-F]{8}-[0-9A-F]{4}-5[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$"
           },
           gt?: {exclusiveMinimum: IDENTITY},
-          gteq?: {mininum: IDENTITY},
+          gteq?: {minimum: IDENTITY},
           lt?: {exclusiveMaximum: IDENTITY},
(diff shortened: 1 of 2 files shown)


solnic (approved):

solnic:
Thank you! ❤️ 🙇🏻

PR #408, opened 2022-06-21
Support anyOf composition in JSON schema output

robhanlon22:
Hi! It's been awhile 😄 This implements `visit_or` within Dry::Schema::JSONSchema::SchemaCompiler so that composition of predicates and schemata can be converted to JSON schemas.


solnic (approved):
👏🏻 🙇🏻

PR #412, opened 2022-06-28
Allow nested filled and value macro usage

robhanlon22:
All of the macros have access to the macro DSL. With a few small tweaks,
it's possible to make the macros apply to the values themselves,
allowing things like nested arrays to be validated without having to use
predicate logic.


solnic (approved):
NICE

PR #414, opened 2022-07-06
[wip] make everything except extensions autoloadable

solnic:


solnic:
@gmcgibbon @rafaelfranca 👋🏻 FYI this is a follow-up after your original autoload work but it uses zeitwerk. I need to tweak zeitwerk loader so that it knows how to deal with extension files (or just move files around a bit). Overall, this looks very promising. Notice that it uses dry-core/configurable/types from main here because those are now autoloadable via zeitwerk as well. Once this one's working too, I'll make dry-validation autoloadable and then we could test it all out 😅 Assuming everything works fine, I'll release all of the things and we should end up with a major boot time speed up ⚡

rafaelfranca:
🎉

solnic:
I closed this this since it was done in other PRs eventually

PR #416, opened 2022-07-07
Fix for filled + constrained constructor type

robhanlon22:
Since extract_type_spec was receiving *args, the splat could possibly
include something interpreted as a keyword arg (at least on Ruby 2.7.x).
Make extract_type_spec take an unambiguous array.


solnic:
Please rebase this as well 😄

robhanlon22:
@solnic done and ready!

solnic (approved):
Ohhh thanks!!!

PR #420, opened 2022-08-10
Or::SinglePath and Or::MultPath robustness fixes

robhanlon22:
It's possible for left and right to be an array of messages when
creating an Or::SinglePath, and it's possible for an Or::MultiPath to
contain an Or::SinglePath. Adds the following fixes:

* Turn left and right into flattened arrays when constructing
  Or::SinglePaths, then extract one of the paths (they should all be
  the same). Update all methods to handle arrays of messages.
* Implement to_or on Or::SinglePath, as they may end up in
  Or::MultiPaths.
* Simplify Or::MultiPath.handler to just use a case statement. The
  hash-based dispatch was unnecessarily complex.

### DIFF
--- a/lib/dry/schema/message/or/multi_path.rb
+++ b/lib/dry/schema/message/or/multi_path.rb
@@ -34,17 +34,17 @@ module Dry
             end
           end
 
-          # @api private
-          def self.handler(message)
-            handlers.find { |k,| message.is_a?(k) }&.last
-          end
+          MULTI_PATH_HANDLER = -> { _1 }
+          MESSAGE_ARRAY_HANDLER = -> { MessageArray.new(_1) }
 
           # @api private
-          private_class_method def self.handlers
-            @handlers ||= {
-              self => :itself.to_proc,
-              Array => MessageArray.method(:new)
-            }.freeze
+          def self.handler(message)
+            case message
+            when self
+              MULTI_PATH_HANDLER
+            when Array
+              MESSAGE_ARRAY_HANDLER
+            end
           end
 
           # @api public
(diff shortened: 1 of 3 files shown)


flash-gordon:
> Simplify Or::MultiPath.handler to just use a case statement. The hash-based dispatch was unnecessarily complex.

Now `:itself.to_proc` and `.method(:new)` are called on every call of `.handler`, their results should be constants instead.

robhanlon22:
I was interested in seeing how Ruby handles this, looks like `Keyword#to_proc` always returns the same object:

```
irb(main):002:0> :itself.to_proc.equal?(:itself.to_proc)
=> true
```

But `Object#method` always returns new objects:

```
irb(main):004:0> Object.method(:new).equal?(Object.method(:new))
=> false
```

TIL 😁

flash-gordon:
> TIL 😁

I wouldn't count on Ruby too much in these aspects. They tend to tighten things a bit but it takes time. I'm pretty sure there's not much practical sense in mutable `Method` instances, but everything is mutable by default except for symbols and integers.

lib/dry/schema/message/or/multi_path.rb
```diff
@@ -34,17 +34,17 @@ def to_h
             end
           end
 
-          # @api private
-          def self.handler(message)
-            handlers.find { |k,| message.is_a?(k) }&.last
-          end
+          MULTI_PATH_HANDLER = -> { _1 }
```
solnic, line 37: We have `IDENTITY` function defined in `Dry::Core::Constants` which is supposed to be available here, please check.

solnic (approved):
Approved albeit one minor thing is left to be addressed

PR #421, opened 2022-08-16
Fix value coercion for composed schemas

robhanlon22:
Adds DSL#strict_type_schema. This is used when composing schemas within
macros. Making them strict improves composition of schemas, especially
when using the | operator. Before, all of the composed schemas were lax,
and if a subschema would pass a lax check, then values would be
eliminated from the coerced output that are necessary for later schema
validation to succeed.

Also fixes up metadata when creating the types map for a schema to not
always set required to false. This was also causing incorrect elision of
required keys when composing schemas.

### DIFF
--- a/lib/dry/schema/dsl.rb
+++ b/lib/dry/schema/dsl.rb
@@ -288,13 +288,20 @@ module Dry
 
       # Return type schema used by the value coercer
       #
-      # @return [Dry::Types::Safe]
+      # @return [Dry::Types::Lax]
       #
       # @api private
       def type_schema
-        our_schema = type_registry["hash"].schema(types).lax
-        schemas = [*parents.map(&:type_schema), our_schema]
-        schemas.inject { |result, schema| result.schema(schema.to_a) }
+        strict_type_schema.lax
+      end
+
+      # Return type schema used when composing subschemas
+      #
+      # @return [Dry::Types::Schema]
+      #
+      # @api private
+      def strict_type_schema
+        type_registry["hash"].schema(types)
       end
 
       # Return a new DSL instance using the same processor type
@@ -316,7 +323,7 @@ module Dry
       # @api private
       def set_type(name, spec)
         type = resolve_type(spec)
-        meta = {required: false, maybe: type.optional?}
+        meta = {required: true, maybe: type.optional?}
 
         @types[name] = type.meta(meta)
       end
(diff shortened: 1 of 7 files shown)


robhanlon22:
@solnic this would be great to get into 1.10.1! :)

lib/dry/schema/dsl.rb
```diff
@@ -288,13 +288,20 @@ def parent
...
+        strict_type_schema.lax
+      end
+
+      # Return type schema used when composing subschemas
+      #
+      # @return [Dry::Types::Schema]
+      #
+      # @api private
+      def strict_type_schema
+        type_registry["hash"].schema(types)
```
robhanlon22, line 304: No longer necessary to merge parent `type_schema`s here because this is handled by `DSL#types` now.

lib/dry/schema/types_merger.rb
```diff
@@ -63,35 +61,42 @@ def handle_ordered
...
-                op_class,
-                unwrapped_old.name_key_map,
-                unwrapped_new.name_key_map
-              )
-            )
+            merge_schemas(unwrapped_old, unwrapped_new)
           in [Dry::Types::AnyClass, _] | [Dry::Types::Hash, Dry::Types::Schema]
             unwrapped_new
-          in [Dry::Types::Hash, Dry::Types::Schema] | [_, Dry::Types::AnyClass]
+          in [Dry::Types::Schema, Dry::Types::Hash] | [_, Dry::Types::AnyClass]
```
robhanlon22, line 74: This was a bug :)

lib/dry/schema/dsl.rb
```diff
@@ -316,7 +323,7 @@ def new(**options, &block)
       # @api private
       def set_type(name, spec)
         type = resolve_type(spec)
-        meta = {required: false, maybe: type.optional?}
```
robhanlon22, line 319: By setting `required` to `true` here, we can be assured that each type will be able to be used correctly for any downstream value coercion. The value coercion's use of `type_schema` applies `lax` to the underlying types. Without this change, all types are somewhat lax by default which, in the case of operator composition, leads to incorrect removal of keys, as the semi-lax types will cause non-matching type cases to succeed.

Since none of the tests failed with this change, I feel pretty safe about this, but are there some cases where this might cause issues?

PR #419, opened 2022-08-03
Custom sum types fail in main

flash-gordon:
The spec in this PR works in 1.9.3 but fails in our main branch with the following backtrace
```
  1) Registering custom types class-based definition maybe decimal coerces the type
     Failure/Error: types.registered?(key) ? types[key] : types[name.to_s]
     
     Dry::Container::KeyError:
       key not found: "decimal?"
       Did you mean?  "decimal"
     # /Users/gordon/.rvm/gems/ruby-3.1.2@dry-schema/gems/dry-container-0.10.1/lib/dry/container/resolver.rb:31:in `block in call'
     # /Users/gordon/.rvm/gems/ruby-3.1.2@dry-schema/gems/dry-container-0.10.1/lib/dry/container/resolver.rb:27:in `fetch'
     # /Users/gordon/.rvm/gems/ruby-3.1.2@dry-schema/gems/dry-container-0.10.1/lib/dry/container/resolver.rb:27:in `call'
     # /Users/gordon/.rvm/gems/ruby-3.1.2@dry-schema/gems/dry-container-0.10.1/lib/dry/container/mixin.rb:171:in `resolve'
     # /Users/gordon/.rvm/gems/ruby-3.1.2@dry-schema/gems/dry-container-0.10.1/lib/dry/container/mixin.rb:184:in `[]'
     # /Users/gordon/.rvm/gems/ruby-3.1.2@dry-schema/bundler/gems/dry-types-8713a790a94f/lib/dry/types.rb:97:in `block in []'
     # /Users/gordon/.rvm/gems/ruby-3.1.2@dry-schema/gems/concurrent-ruby-1.1.10/lib/concurrent-ruby/concurrent/map.rb:203:in `block in fetch_or_store'
     # /Users/gordon/.rvm/gems/ruby-3.1.2@dry-schema/gems/concurrent-ruby-1.1.10/lib/concurrent-ruby/concurrent/map.rb:182:in `fetch'
     # /Users/gordon/.rvm/gems/ruby-3.1.2@dry-schema/gems/concurrent-ruby-1.1.10/lib/concurrent-ruby/concurrent/map.rb:202:in `fetch_or_store'
     # /Users/gordon/.rvm/gems/ruby-3.1.2@dry-schema/bundler/gems/dry-types-8713a790a94f/lib/dry/types.rb:88:in `[]'
     # ./lib/dry/schema/type_registry.rb:40:in `[]'
```
[...]

PR #425, opened 2022-09-08
Custom type fixes

### LINKED ISSUE #423: Upgrade past 1.9.3 introduces KeyError to Production App
## Describe the bug

When attempting to programmatically add a custom type to a schema, we are receiving the following error:
```
Dry::Container::KeyError:
  key not found: "{:type?=>Types::CalendarDate}"
```

I believe this was introduced when `extract_type_spec` was wrapped around the `value` and `filled` internals within `Schema::Macros::DSL`.

## To Reproduce

The minimal amount of code to reproduce using RSpec tests:
```
# frozen_string_literal: true

require 'dry-schema'
require 'dry-container'
require 'dry-types'

module Types
  include Dry.Types()

  class CalendarDate < ::Date
    def to_json(*_a)
      strftime("--%m-%d")
    end
    alias_method :as_json, :to_json

    def self.parse(date)
      mon, mday = Date._iso8601(date).values_at(:mon, :mday)
      raise ArgumentError, "invalid ISO8601 calendar day string, expected format \"--MM-DD\"" unless mon && mday

      new(2000, mon, mday)
    end
  end

  def self.to_calendar_date(input, &block)
    CalendarDate.parse(input)
  end

  CalendarDay = Nominal(CalendarDate).constructor(method(:to_calendar_date))

  Container = Dry::Schema::TypeContainer.new
  Container.register(:calendar_day) { Types::CalendarDay }
end

Schema = Dry::Schema.JSON do
  config.types = Types::Container
  optional(:date).maybe(:calendar_day)
end

RSpec.describe Schema do
  let!(:data) { { "date" => "--02-09" } }
  let(:result) { described_class.(data) }

  context "call" do
    it "is expected to succeed" do
      expect(result).to be_a_success
    end
  end
end
```

## Expected behavior

The above test passes on version `1.9.3` but now fails on versions `1.10.1` and `1.10.2`

I would expect this to pass as written in 1.10.2.

## My environment

- Affects my production application: YES
- Ruby version: 3.0.4

robhanlon22:
Appends the Value macro directly instead of calling .value and re-extracting the type spec.

For consistency, hoist maybe to the DSL as well so it can be used inside of blocks, just like filled.

Incorporates a variant of @ADStrovers repro spec from #423.

Resolves #419.
Resolves #423.

### DIFF
--- a/lib/dry/schema/macros/dsl.rb
+++ b/lib/dry/schema/macros/dsl.rb
@@ -83,6 +83,24 @@ module Dry
           end
         end
 
+        # Set type specification and predicates for a maybe value
+        #
+        # @example
+        #   required(:name).maybe(:string)
+        #
+        # @see Macros::Key#value
+        #
+        # @return [Macros::Key]
+        #
+        # @api public
+        def maybe(*args, **opts, &block)
+          extract_type_spec(args, nullable: true) do |*predicates, type_spec:, type_rule:|
+            append_macro(Macros::Maybe) do |macro|
+              macro.call(*predicates, type_spec: type_spec, type_rule: type_rule, **opts, &block)
+            end
+          end
+        end
+
         # Specify a nested hash without enforced `hash?` type-check
         #
         # This is a simpler building block than `hash` macro, use it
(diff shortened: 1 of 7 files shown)


flash-gordon (approved):
👍👍👍

solnic:
Thanks for taking care of this. I'm gonna release it today/tomorrow.

PR #431, opened 2022-09-22
Update config in configure blocks only

timriley:
Update all usages of dry-configurable-provided `config` so that any changes are made inside `configure` blocks.

This prepares for the changes in https://github.com/dry-rb/dry-configurable/pull/140.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,10 +9,7 @@ gemspec
 # if ENV["DRY_CONFIGURABLE_FROM_MAIN"].eql?("true")
 #   gem "dry-configurable", github: "dry-rb/dry-configurable", branch: "main"
 # end
-gem "dry-configurable",
-  github: "dry-rb/dry-configurable",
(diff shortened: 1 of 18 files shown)


solnic:
Oh but this means it should go to `dry-schema` 2.0.0 :/ Is it possible to actually maintain previous behavior?

timriley:
Just noting for the record that we discussed over in https://github.com/dry-rb/dry-configurable/pull/140 that we would:

> make the type spec change (making them mandatory) and push 2.0 along with the configure change

flash-gordon:
Should we plan zeitwerk for 2.0 as well? I guess we should.

flash-gordon:
For the record, I've been running zeitwerk for 2+ months on production without problems

solnic:
> Should we plan zeitwerk for 2.0 as well? I guess we should.

I say yes. Most of the work has been completed. It's a matter of releasing stuff and fixing any potential issues (😅).

Issue #304, opened 2020-07-13
If using custom top_namespace, custom namespace not work


moreta:
**Describe the bug**

I found this  bug when using dry-validation.
dry-validation's default top_namespace is 'dry_validation'.
Since, When using custom namespace with dry_validation, custom namespace key name not founded

**To Reproduce**

```ruby
# test.yml
# en:
#   dry_validation:
#     post:
#       rules:
#         post_body: 'My (dry_validation) Post body'
#   dry_schema:
#     post:
#       rules:
#         post_body: 'My (dry_schema) Post body'
RSpec.describe "Namespaced messages with Custom Top namespaced" do

  # Fail case
  context "namespace with Custom Top namespaced schema" do
    let(:post_schema) do
      Dry::Schema.Params do
        config.messages.load_paths << "#{SPEC_ROOT}/fixtures/locales/test.yml"
        config.messages.top_namespace = 'dry_validation'
        config.messages.namespace = :post

        required(:post_body).filled(:string)
      end
    end

    it "uses namespaced full messages" do
      result = post_schema.call(post_body: "")
      expect(result.errors(full: true).to_h[:post_body]).to eql(["My (dry_validation) Post body must be filled"])
    end
  end

  # Success case
  context "namespace with Custom Top namespaced schema" do
    let(:post_schema) do
      Dry::Schema.Params do
        config.messages.load_paths << "#{SPEC_ROOT}/fixtures/locales/test.yml"
        config.messages.top_namespace = 'dry_schema'
        config.messages.namespace = :post

        required(:post_body).filled(:string)
      end
    end

    it "uses namespaced full messages" do
      result = post_schema.call(post_body: "")
      expect(result.errors(full: true).to_h[:post_body]).to eql(["My (dry_schema) Post body must be filled"])
    end
  end
end
```

**Expected behavior**

Above unit test's "Fail case" should be success.

**Your environment**

- Ruby version: ruby-2.6.5
- OS: macOS 10.15.5

PR #432, opened 2022-09-29
Fix issues with rule name when top_namespace and namespace passed

RudskikhIvan:
Fix for the issue [#304](https://github.com/dry-rb/dry-schema/issues/304)


solnic:
Good catch, thank you! ❤️

Issue #438, opened 2022-11-02
Schema definition regression after 1.9.3


pyromaniac:
Hello there folks!

## Describe the bug

There was a weird schema regression after 1.9.3.

## To Reproduce

```ruby
s = Dry::Schema.Params do
  optional(:review).maybe(:hash) do
    required(:rating).filled(:decimal)
    optional(:comment).maybe(:string)
  end
end
s.call(review: { rating: 3 })
#=> #<Dry::Schema::Result{:review=>{:rating=>3}} errors={:review=>{:rating=>["must be a decimal"]}} path=[]>
```

Unfortunately, unable to debug it deeper but `.maybe(:hash)` is important and `optional(:comment).maybe(:string)` presence is also important. It could be some leakage in the schema definition.

## Expected behavior

In 1.9.3 it works as expected passing the validation.

## My environment

- Affects my production application: **YES**
- Ruby version: 2.7.6
- OS: OS X/Linux on CI

Big thank you in advance!

flash-gordon:
At the first glance this is connected https://github.com/dry-rb/dry-validation/issues/718 😅

pyromaniac:
It indeed looks the same issue. I also had a bigger schema and managed to reduce the example to the one I submitted and it looks very similar to https://github.com/dry-rb/dry-validation/issues/718#issuecomment-1284124753

solnic:
I'll look into this. Most likely it's the same story as with #437.

PR #439, opened 2022-11-14
Change required metadata back to true

robhanlon22:
This had some unintended effects as far as value coercion goes. At some point, this was causing issues with TypesMerger, but now it appears that it does not! Great!

Resolves dry-rb/dry-validation#718
Resolves dry-rb/dry-schema#438

### DIFF
--- a/spec/integration/issues_spec.rb
+++ b/spec/integration/issues_spec.rb
@@ -1,0 +1,100 @@
+# frozen_string_literal: true
+
+RSpec.describe "GitHub issues" do
+  describe "dry-rb/dry-validation#718" do
+    let(:schema) do
+      Dry::Schema.Params do
+        optional(:results).maybe(:hash) do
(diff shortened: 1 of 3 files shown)


spec/integration/issues_spec.rb
```diff
@@ -0,0 +1,100 @@
+# frozen_string_literal: true
+
+RSpec.describe "GitHub issues" do
+  describe "dry-rb/dry-validation#718" do
```
flash-gordon, line 4: I have git blame for that. I'm not against having some real-world use cases but we need more targeted specs as well. The problem with specs like here is if you work on something and it breaks it can be really hard to find the cause

solnic:
Thanks! I'm gonna work on merging this and releasing stuff just before Hanami 2.0 release. For now it needs to wait a little bit more.

PR #440, opened 2022-11-20
Make the yaml cache fully thread-safe

mensfeld:
```ruby
Concurrent::Map.new { |h, k| h[k] = Concurrent::Map.new }
```

is prone to a race condition, thus the cache won't be fully thread safe if definitions are parsed in multiple threads.

This PR fixes that.

Ref: https://github.com/ruby-concurrency/concurrent-ruby/issues/970

### DIFF
--- a/lib/dry/schema/messages/yaml.rb
+++ b/lib/dry/schema/messages/yaml.rb
@@ -164,7 +164,6 @@ module Dry
 
         # @api private
         def evaluation_context(key, options)
-          p key
           cache.fetch_or_store(get(key, options).fetch(:text)) do |input|
             tokens = input.scan(TOKEN_REGEXP).flatten(1).map(&:to_sym).to_set
             text = input.gsub("%", "#")
--- a/lib/dry/schema/messages/yaml.rb
+++ b/lib/dry/schema/messages/yaml.rb
@@ -70,7 +70,9 @@ module Dry
 
         # @api private
         def self.cache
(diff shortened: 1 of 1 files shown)


lib/dry/schema/messages/yaml.rb
```diff
@@ -162,6 +164,7 @@ def interpolate(key, options, **data)
 
         # @api private
         def evaluation_context(key, options)
+          p key
```
solnic, line 167: I guess this should be removed

solnic (changes requested):

mensfeld:
specs dont seem to be related to this change

solnic (changes requested):
I'm afraid spec failures are indeed caused by the fix, seems like cache isn't cleared properly between spec runs now.

mensfeld:
Got it. WIll look at this today.

flash-gordon:
main is green, merging

granthusbands:
@mensfeld  You missed the `||=` that you've fixed in pull requests for other projects.

mensfeld:
@granthusbands I did not miss it. I explicitly do not do it here to make the change small. I pointed out in the article that I wanted to keep the scope small because I do not know how given libs operate.

>  that you've fixed in pull requests for other projects.

Not exactly the same. In most cases, I fixed the internal `||=`, not the external. Nonetheless you are right and feel free to create a PR :) @solnic should be happy

Issue #441, opened 2022-11-21
Warning about pattern matching on Ruby 2.7


r7kamura:
## Describe the bug

As stated in the gemspec, this gem should support Ruby 2.7 or higher.

```ruby
spec.required_ruby_version = ">= 2.7.0"
```

- https://github.com/dry-rb/dry-schema/blob/v1.12.0/dry-schema.gemspec#L34

Even so, pattern matching is used in the code, which should not have been formally introduced in Ruby 2.7, and which is still in the experimental phase.

- https://github.com/dry-rb/dry-schema/blob/v1.12.0/lib/dry/schema/types_merger.rb#L71-L80

As a result, various codes that use this gem (e.g., Rails apps with config gem) are now generating the following warning when run under Ruby 2.7.

```
$ ruby -v
ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-linux]
$ bin/rails c
/usr/local/bundle/gems/dry-schema-1.10.6/lib/dry/schema/types_merger.rb:71: warning: Pattern matching is experimental, and the behavior may change in future versions of Ruby!
Loading development environment (Rails 7.0.3.1)
irb(main):001:0> 
```

This pattern matching code was added in the following Pull Request and appears to have been introduced in v1.10.0.

- https://github.com/dry-rb/dry-schema/pull/415

## To Reproduce

The easiest way to reproduce this is to `require 'dry/schema/types_merger'` on Ruby 2.7.

```
$ ruby -v
ruby 2.7.6p219 (2022-04-12 revision c9c2245c0a) [x86_64-linux]
$ bundle exec ruby -I lib -e "require 'dry/schema/types_merger'"
/home/r7kamura/ghq/github.com/dry-rb/dry-schema/lib/dry/schema/types_merger.rb:71: warning: Pattern matching is experimental, and the behavior may change in future versions of Ruby!
```

## Expected behavior

If Ruby 2.7 is supported, at least the implementation in the lib directory should not use pattern matching so that no warnings are generated at runtime. And if we use pattern matching, we should drop support for Ruby 2.7.

## My environment

- Affects my production application: YES
- Ruby version: 2.7.2, 2.7.6
- OS: Ubuntu 20.04

PR #442, opened 2022-11-21
Replace case-in pattern-matching with normal case-when statements

r7kamura:
Fix https://github.com/dry-rb/dry-schema/issues/441


r7kamura:
I made a simple change from case-in to case-when for now, but RuboCop will probably be upset with this code about its Metrics, so I'll have to add some more refactoring.

```
$ bundle exec rubocop lib/dry/schema/types_merger.rb
Inspecting 1 file
C

Offenses:

lib/dry/schema/types_merger.rb:70:9: C: Metrics/PerceivedComplexity: Perceived complexity for merge_unwrapped_types is too high. [10/8]
        def merge_unwrapped_types(unwrapped_old, unwrapped_new) ...
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

1 file inspected, 1 offense detected

Tip: Based on detected gems, the following RuboCop extension libraries might be helpful:
  * rubocop-rake (https://rubygems.org/gems/rubocop-rake)
  * rubocop-rspec (https://rubygems.org/gems/rubocop-rspec)

You can opt out of this message by adding the following to your config (see https://docs.rubocop.org/rubocop/extensions.html#extension-suggestions for more options):
  AllCops:
    SuggestExtensions: false
```

flash-gordon:
Please replace it with conditional `class_eval` for ruby 2.7/3+. We'll drop support for 2.7 soon anyway so it'll be easier to go with the PM version later

solnic:
Thanks for tackling this. It'd be cleaner to just have two modules and include them conditionally though 😬

Issue #446, opened 2022-12-02
Sum hash types allow any hash to pass coercion


unkmas:
## Describe the bug

Schema containing sum hash types allow any hash to pass coercion.

## To Reproduce

```ruby
require 'dry-schema'

module Types
  include Dry.Types()
end

FooType = Types::Hash.schema(foo: Types::String)
BarType = Types::Hash.schema(bar: Types::String)

FooBarType = FooType | BarType

FooBarSchema = Dry::Schema.Params do
  required(:foobar).filled(FooBarType)
end

data = {foobar: {something_wrong: true}}

FooBarSchema.call(data)
# => #<Dry::Schema::Result{:foobar=>{:something_wrong=>true}} errors={} path=[]>

# Type itself works fine
FooBarType[data[:foobar]]
# => `block in resolve_missing_keys': :bar is missing in Hash input (Dry::Types::MissingKeyError)

# Hash schemas without sums also works fine:
FooSchema = Dry::Schema.Params do
  required(:foobar).filled(FooType)
end

FooSchema.call(data)
#=> #<Dry::Schema::Result{:foobar=>{}} errors={:foobar=>["must be filled"]} path=[]>
```

## Expected behavior

Coercion should happen for sum hash types.

## My environment

- Ruby version: `ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [arm64-darwin21]`

solnic:
This is now fixed in 1.13.1

PR #457, opened 2023-04-06
Support sum types of hashes

segiddins:
Fixes https://github.com/dry-rb/dry-schema/issues/446, which I believe is necessary because of https://github.com/dry-rb/dry-types/pull/395


solnic:
Thank you! ❤️

PR #458, opened 2023-04-07
Support sum types of hashes in arrays

segiddins:


segiddins:
Interestingly, `bundle exec rubocop` isn't showing any failures locally for me

PR #462, opened 2023-06-20
Fix type of objects nested under arrays. Fixes #400

### LINKED ISSUE #400: Deeply nested objects have the wrong JSON schema type
## Describe the bug

Given these two dry schemas:

```ruby
A = Dry::Schema.JSON do 
  required(:attributes).maybe(:hash) do
    required(:name).filled(:string, min_size?: 1)
  end
end

B = Dry::Schema.JSON do 
  required(:data).array(A)
end
```

I would expect to se a JSON schema like this:

```json
{
  "type": "object",
  "properties": {
    "data": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "attributes": {
            "type": "object",
            "properties": {
              "name": {
                "type": "string",
                "minLength": 1
              }
            },
            "required": [
              "name"
            ]
          }
        },
        "required": [
          "attributes"
        ]
      }
    }
  },
  "required": [
    "data"
  ]
}
```

But, in fact, I see JSON like this:

```json
{
  "type": "object",
  "properties": {
    "data": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "attributes": {
            "type": "array",
            "items": {
              "type": "object",
              "properties": {
                "name": {
                  "type": "string",
                  "minLength": 1
                }
              },
              "required": []
            }
          }
        },
        "required": [
          "attributes"
        ]
      }
    }
  },
  "required": [
    "data"
  ]
}
```

There are two, principle differences:

1. I would expect the `attributes` to be an object, since it's defined like this: `required(:attributes).maybe(:hash)`. It is, in fact, an array.
2. I would expect `attributes` to require the `name` since it's defined like this `required(:name).filled(:string, min_size?: 1)`. It is, in fact, not required.

- Ruby version: 3.1.1
- OS: Mac OS Monteray (M1)

tomgi:
Fix for https://github.com/dry-rb/dry-schema/issues/400

Objects nested under arrays are not neccesailry arrays themselves - the `:member` option shouldn't be passed down the AST.

### DIFF
--- a/lib/dry/schema/extensions/json_schema/schema_compiler.rb
+++ b/lib/dry/schema/extensions/json_schema/schema_compiler.rb
@@ -90,7 +90,7 @@ module Dry
         def visit_set(node, opts = EMPTY_HASH)
           target = (key = opts[:key]) ? self.class.new(loose: loose?) : self
 
-          node.map { |child| target.visit(child, opts) }
+          node.map { |child| target.visit(child, opts.except(:member)) }
 
           return unless key
 
--- a/spec/extensions/json_schema/schema_spec.rb
+++ b/spec/extensions/json_schema/schema_spec.rb
@@ -27,6 +27,10 @@ RSpec.describe Dry::Schema::JSON, "#json_schema" do
 
         required(:roles).array(:hash) do
           required(:name).value(:string, min_size?: 12, max_size?: 36)
+
+          required(:metadata).hash do
+            required(:assigned_at).value(:time)
+          end
         end
 
         optional(:address).hash do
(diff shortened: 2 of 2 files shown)


cpgo:
Hey @tomgi just asked about this issue on the zulip chat.
Tried monkey patching your solution on my project and it seems to be working perfectly.

@solnic is there any plans to merge this PR?

shauns:
Following up on this -- @solnic would merging it be possible? The bug is still in place and this fix is a good one.

solnic:
I'm sorry it took so long, I'll push a release, thanks for addressing it

lukelex:
Awesome work. Thanks for dedicating your time to fix this :heart:

PR #471, opened 2023-11-07
Fix info extension for maybe macro

santiagodoldan:
Fixes info extension for maybe macro, it introduces `nullable` flag to represent whether a key is nullable or not.

```ruby
    schema = Dry::Schema.JSON do
      optional(:age).maybe(:integer)
    end
```
**BEFORE**
```
irb(main):030> schema.info
/usr/local/bundle/ruby/3.2.0/gems/dry-schema-1.13.3/lib/dry/schema/extensions/info/schema_compiler.rb:46:in `public_send': undefined method `visit_not' for #<Dry::Schema::Info::SchemaCompiler:0x00007fc01e0ce780 @keys={:email=>{:required=>true, :type=>"string"}, :age=>{:required=>false}}> (NoMethodError)

          public_send(:"visit_#{meth}", rest, opts)
          ^^^^^^^^^^^
Did you mean?  visit_and
               visit_set
               visit_key
```
**AFTER**
```
irb(main):156> schema.info
=> {:keys=>{:email=>{:required=>true, :nullable=>false, :type=>"string"}, :age=>{:required=>false, :nullable=>true, :type=>"integer"}}}
```

### DIFF
--- a/lib/dry/schema/extensions/info/schema_compiler.rb
+++ b/lib/dry/schema/extensions/info/schema_compiler.rb
@@ -71,7 +71,7 @@ module Dry
         # @api private
         def visit_implication(node, opts = EMPTY_HASH)
           node.each do |el|
-            visit(el, opts.merge(required: false))
+            visit(el, opts.merge(required: false, nullable: false))
           end
         end
 
@@ -83,7 +83,14 @@ module Dry
         # @api private
         def visit_key(node, opts = EMPTY_HASH)
           name, rest = node
-          visit(rest, opts.merge(key: name, required: true))
+          visit(rest, opts.merge(key: name, required: true, nullable: false))
+        end
+
+        # @api private
+        def visit_not(_node, opts = EMPTY_HASH)
+          key = opts[:key]
+
+          keys[key][:nullable] = true
         end
 
         # @api private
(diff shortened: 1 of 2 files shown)


santiagodoldan:
@solnic could you let me know if my changes make sense? it seems you implemented this class and it would be nice to understand if my changes are aligned with your initial idea, thanks 💯

santiagodoldan:
@solnic Let me know if I need to apply any changes here, I'm not sure if what I did is aligned with your initial approach.

lib/dry/schema/extensions/info/schema_compiler.rb
```diff
@@ -83,7 +83,14 @@ def visit_each(node, opts = EMPTY_HASH)
...
-          visit(rest, opts.merge(key: name, required: true))
+          visit(rest, opts.merge(key: name, required: true, nullable: false))
+        end
+
+        # @api private
+        def visit_not(_node, opts = EMPTY_HASH)
+          key = opts[:key]
+
+          keys[key][:nullable] = true
         end
```
flash-gordon, line 94: Sorry, I don't have time for a deeper look but this analysis seems very shallow. You cannot deduce a key is nullable from merely using `not`, this is too general. Of course, the tests pass but it's only because there are too few of them. I think you ought to analyze in `visit_predicate`, this is the place where you can detect if it was the maybe macro.

santiagodoldan: Sure, will do that 💯

PR #472, opened 2023-11-16
Update macros.html.md

contentfree:
Remove the `hash?` predicate from the "expands to" comment for macros/#schema


flash-gordon:
```ruby
Dry::Schema.Params do
  required(:tags).filled(:hash).schema do
  end
end
# <Dry::Schema::Params
#   keys=[
#     {
#       "tags" => ["name"]
#     }
#   ]
#   rules={
#     :tags => "key?(:tags) AND key[tags](
#       filled? AND hash? AND set(
#         key?(:name) AND key[name](
#           filled? AND str?
#         )
#       )
#     )"
#   }
# >
```
It certainly has the `hash?` predicate inside

PR #482, opened 2024-10-21
update README with instructions on sum types

timhwang21:
I had to look into specs to discover that this behavior was supported:

https://github.com/dry-rb/dry-schema/blob/main/spec/integration/params/macros/value_spec.rb#L31

I feel this is a very useful pattern, and that making it more discoverable will be more user-friendly.


solnic:
Thanks!

PR #483, opened 2024-12-25
Improve support for Sets in `included_in?` and `excluded_from?` predicates

eriklovmo:
This pull request enhances support for using [Set](https://docs.ruby-lang.org/en/master/Set.html) with the `included_in?` and `excluded_from?` predicates:

* Update compiled error messages for sets. With this change, messages like `"must be one of: 1, 3, 5"` are produced instead of `"must be one of: #<Set: {1, 3, 5}>"`. This makes the error clearer, especially when exposed to end users.
* Add tests to demonstrate that Set is supported, as it responds to `#include?`.

Using sets is arguably more natural and efficient for membership checks, so explicitly supporting it makes sense.

### DIFF
--- a/lib/dry/schema/message_compiler.rb
+++ b/lib/dry/schema/message_compiler.rb
@@ -209,7 +209,7 @@ module Dry
       def message_tokens(args)
         tokens = args.each_with_object({}) do |arg, hash|
           case arg[1]
-          when Array
+          when Array, Set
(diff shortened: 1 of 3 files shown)


flash-gordon:
Thanks, I merged this in https://github.com/dry-rb/dry-schema/commit/15d26d43905a4316887f15769faf92e7052495df

PR #485, opened 2025-02-05
Switch to json_schemer from json-schema

michaelherold:
json-schema is unmaintained and only supports up to draft-06 of JSON Schema. That is three versions older than the latest specification. json_schemer is supported and supports all versions of JSON Schema, including 2020-12, which is the latest one.

This change swaps one dependency for another in order to support newer versions.


flash-gordon:
Thank you!

PR #486, opened 2025-02-19
FEAT: add open api schema extension

rreckonerr:
This PR refactors the JSON schema compiler logic by extracting it into a base class, enabling the new OpenAPI schema extension to reuse and extend its functionality.

In the process of migrating our Rails request tests to rswag, I converted our Dry Schemas directly into OpenAPI schema. Since Dry Schema already had JSON Schema extension support, this approach was straightforward to implement.

*Note: Although the contribution guidelines suggest linking to a reported issue, I wasn't able to create an issue for this functionality without using a bug report template.*

### DIFF
--- a/docsite/source/extensions/open_api_schema.html.md
+++ b/docsite/source/extensions/open_api_schema.html.md
@@ -1,0 +1,44 @@
+---
+title: OpenAPI Schema
+layout: gem-single
+name: dry-schema
+---
+
+The `:open_api_schema` extension allows you to generate a valid [OpenAPI Schema](https://swagger.io/specification/v3/) from a `Dry::Schema`. This makes it straightforward to leverage tools like Swagger, which is popular for API documentation and testing.
+
+```ruby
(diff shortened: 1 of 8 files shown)


docsite/source/extensions/open_api_schema.html.md
```diff
@@ -0,0 +1,44 @@
...
+# {
+#   type: "object",
+#   properties: {
+#     email: {
+#       type: "string",
+#       minLength: 8
+#     },
+#     favorite_color: {
+#       type: "string",
+#       minLength: 1,
```
rreckonerr, line 28: Hmm, this seems to be not necessary for "filled" enums, will add more tests for cases like this and adjust the implementation

rreckonerr: We should also add an additional `null` enum along with the `nullable: true` flag for optional [enums](https://swagger.io/docs/specification/v3_0/data-models/enums/).

PR #491, opened 2025-06-30
Fix: Allow use of symbol for configuration `top_namespace`

unused:
Using a `top_namespace` in configuration that is a symbol fails with further processing where a string is expected. Relying on a conversion in settings constructor prevents an unexpected error.

> What error is happening?

Given a symbol is used for `top_namespace`, it results in the following error:

```
Namespaced messages with top namespace allows symbol for top namespace
     Failure/Error: data.transform_keys { _1.gsub(DEFAULT_MESSAGES_ROOT, config.top_namespace) }

     TypeError:
       no implicit conversion of Symbol into String
```

> [!NOTE]
> Ref: [Related issue 734 in dry-validation](https://github.com/dry-rb/dry-validation/issues/734)


timriley:
Thanks @unused!

Issue #492, opened 2025-08-29
I18n nested text and missing options when using lambda


roksanarothenberger-monterail:
## Describe the bug

When using I18n for translations and returning hash with lambda the `text` the whole hash is put inside it's `text` field, and initial call for `meta` is not sending options to I18n, so if we want to introduce meta data based on those there is nothing there.

## To Reproduce

```ruby
require "dry-schema"
require "i18n"

I18n.backend.store_translations(:en, dry_schema: {
    errors: {
      max_size?: ->(*, **kargs) { {text: "is over limit", max_size: kargs[:num]} }
    }
  }
)

schema = Dry::Schema.Params do
  config.messages.backend = :i18n
  config.messages.load_paths = []

  required(:type).value(:string, max_size?: 1)
end

puts schema.call({type: "abc"}).errors.to_h
```

```
ruby example.rb

{type: [{text: {text: "is over limit", max_size: 1}, max_size: nil}]}
```

## Expected behavior

Message should not be nested and options should be provided to call for meta attributes.

```
ruby example.rb

{type: [{text: "is over limit", max_size: 1}]}
```

## My environment

- Affects my production application: **YES**
- Ruby version: 3.4.5
- OS: Ubuntu 24.04/Alpine 3.22

PR #493, opened 2025-08-29
Fix handling of I18n proc messages when returning hashes

rrothenberger:
Resolves: https://github.com/dry-rb/dry-schema/issues/492 - at least to the best of my ability 😅


spec/integration/messages/i18n_spec.rb
```diff
@@ -203,6 +214,26 @@ def store_errors(**errors)
...
+              }
+            }
+          )
+
+          template, meta = messages[:predicate_with_meta, path: :pages, locale: :en]
+
+          expect(template.(arg: "some arg")).to eql("text some arg")
+          expect(meta).to eq({code: 123})
+        end
+      end
```
rrothenberger, line 236: Initially I wanted to jest drop the logic around `text_key` in  `lib/dry/schema/messages/i18n.rb` - but it was making this possible and they might be people relying on this behavior so I thought it was worth to add test for it.

timriley:
Thank you @rrothenberger! This looks like a very good fix to me :)

PR #501, opened 2025-10-24
Add description support for schema keys

baweaver:
Adds a `.description(text)` method to schema key definitions that stores descriptions in type 
metadata and includes them in JSON schema output.

### Usage

```ruby
  required(:first_name).filled(:string).description("First name of the user")
  optional(:age).filled(:integer).description("Age of the user")
  required(:address).description("The shipping address").hash do
    required(:street).filled(:string).description("Street address")
    optional(:city).filled(:string).description("City name")
  end
end

# Access via type metadata
schema.types[:first_name].meta[:description]
# => "First name of the user"

# Included in JSON schema output
schema.json_schema[:properties][:first_name][:description]
# => "First name of the user"
```

### Implementation

• Descriptions stored in type metadata (`:description` key)
• Works with all macro types (required, optional, filled, maybe, hash, array, etc.)
• Preserved when types are replaced during macro chaining
• JSON schema compiler extracts descriptions from type metadata for both top-level and nested 
keys
• Improved `set_type` to preserve all incoming type metadata, not just schema-specific meta

### Changes

• `lib/dry/schema/macros/dsl.rb` - Added description method
• `lib/dry/schema/dsl.rb` - Updated `set_type` to preserve type metadata
• `lib/dry/schema/extensions/json_schema.rb` - Pass types to compiler
• `lib/dry/schema/extensions/json_schema/schema_compiler.rb` - Extract and include descriptions
• `spec/integration/schema/description_spec.rb` - Test coverage

Backwards compatible with existing schemas.


lib/dry/schema/dsl.rb
```diff
@@ -315,7 +315,13 @@ def new(klass: self.class, **options, &)
       # @api private
       def set_type(name, spec)
         type = resolve_type(spec)
-        meta = {required: false, maybe: type.optional?}
+        meta = type.meta.dup
+        meta[:required] = false
+        meta[:maybe] = type.optional?
+
+        if !meta.key?(:description) && @types[name]&.meta&.key?(:description)
```
baweaver: There has to be a better way to do this one.

lib/dry/schema/extensions/json_schema.rb
```diff
@@ -17,7 +17,7 @@ module SchemaMethods
         #
         # @api public
         def json_schema(loose: false)
-          compiler = SchemaCompiler.new(root: true, loose: loose)
+          compiler = SchemaCompiler.new(root: true, loose: loose, types: types)
```
baweaver: Need to pipe `types` down as those are where most of the meta / description info lives.

lib/dry/schema/extensions/json_schema/schema_compiler.rb
```diff
@@ -88,7 +89,13 @@ def visit(node, opts = EMPTY_HASH)
...
         # @api private
         def visit_set(node, opts = EMPTY_HASH)
-          target = (key = opts[:key]) ? self.class.new(loose: loose?) : self
+          key = opts[:key]
+          target = if key
+                     nested_types = extract_nested_types(@types[key])
+                     self.class.new(loose: loose?, types: nested_types)
+                   else
+                     self
+                   end
```
baweaver: Why RuboCop why?

lib/dry/schema/macros/dsl.rb
```diff
@@ -199,6 +199,19 @@ def custom_type?
...
+        # Set description for the key
+        #
+        # @param [String] text Description text
+        #
+        # @return [Macros::DSL]
+        #
+        # @api public
+        def description(text)
+          type = schema_dsl.types[name]
+          schema_dsl.set_type(name, type.meta(type.meta.merge(description: text)))
```
baweaver: Is there a nicer way to write `type.meta(type.meta.merge(description: text))` ?

lib/dry/schema/dsl.rb
weedySeaDragon: :wave Hello.  I'm randomly jumping in here to comment.  
 I'm reading thru code as I come up to speed on some `dry-rb` things. I was reading thru issues and PRs to see if something's already been addressed or not and was perusing through your PR.)

Forgive me if you've already considered this:
1.  use `||=` (conditional assignment) instead of  `!meta.key?(:description)`
2. could use `dig` **iff** `@types.meta` will always respond to `.dig` (and I'm not entirely sure if it will, looking at the code in `Dry::Types`)
  Using `dig` is a _little_ better because it works for Hashes, Arrays, and Structs, so it is less bound to the underlying object structure:

```ruby
if @types.dig(name, :meta, :description)
  meta[:description] ||= @types[name].meta[:description]
end
```

Using `dig` used to be a bigger improvement before the safe navigation operator (`&.`) was added to Ruby. 
It's still helpful for making things less tightly bound to the underlying implementation. In this case, that's not an issue (since this is creating and _dealing with_ the implementation), so it's more of a stylistic choice.

(It looks like there are few places in `dry-schema` where `.dig` could be used iff the `Meta` module in `dry-types` could ensure that `.meta` would always respond to `dig`.)

timriley (commented):
This is looking good to me! Thank you @baweaver.

I've left just one question for you, which I think would be good to address before merging.

@flash-gordon Do you have any thoughts about this one?

lib/dry/schema/dsl.rb
```diff
@@ -314,10 +314,17 @@ def new(klass: self.class, **options, &)
...
+        new_type = resolve_type(spec)
 
-        @types[name] = type.meta(meta)
+        new_meta = new_type.meta.dup
+        new_meta[:required] = false
+        new_meta[:maybe] = new_type.optional?
+
+        current_meta = @types[name]&.meta
+
+        new_meta[:description] ||= current_meta[:description] if current_meta&.key?(:description)
```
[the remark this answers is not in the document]
baweaver: Modified and generalized to merge a wider range of metadata, added comment.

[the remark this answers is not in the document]
timriley: Thanks @baweaver, that’s very nice!

stefanoc:
Hey what's the status here? I was just looking for this but it seems it was abandoned 😢 
I have a hand-written Zod schema in our Typescript codebase that I'd really like to auto-generate from a Dry::Schema since it is actually more the responsibility of the Ruby codebase, and I really don't want to have to do it the other way around since I like Ruby but Typescript... not so much :-)

timriley:
Thanks for chiming in here, @stefanoc! I'm keen to see this go in. Let's use your use case as input to the design. If you were to generate your Zod schemas, is that happening via JSON Schema-formatted JSON, or would you take some other approach?

stefanoc:
Zod natively supports JSON Schema since 4.0 so that would be the easiest route, but it's still flagged as experimental so who knows... ideally I'd like to go with a direct Ruby-to-Typescript conversion (so I guess something like the JSON schema extension, but which would emit a string instead of a Hash, if that makes sense).

rolftimmermans:
I'm also curious about the status. I'm currently using `dry-struct` for structured outputs from LLMs, but write the JSON schema definitions manually. That's duplication I prefer to avoid.

It would be ideal to define the schema with `dry-schema`, but not having the ability to include a `description` is a blocker. I'm hopeful this feature will be considered... :)

In practice the descriptions can be pretty long multi-line strings though. A fluent API might not be super ergonomical (although definitely better than not having the feature at all).

PR #508, opened 2025-12-03
FIX support for nested arrays with `config.validate_keys = true`

misdoro:
Before this change, it was impossible to validate schema for nested arrays of strings or hashes in params, when `config.validate_keys = true` was set. 

**Before:**
```ruby
  config.validate_keys = true
  required(:matrix).value(:array, size?: 2).each do
    value(:array, size?: 3).each do
      value(:string, size?: 1)
    end
  end
end

errors = schema.call(matrix: [["a", "b", "c"], ["d", "ee", "f"]]).errors

puts errors.to_h.inspect
```
would fail validation saying:
```ruby
{matrix: {0 => {a: ["is not allowed"], b: ["is not allowed"], c: ["is not allowed"]}, 1 => {d: ["is not allowed"], ee: ["is not allowed"], f: ["is not allowed"]}}}
```

**After**

a correct validation error for the string length is reported
```ruby
{:matrix=>{1=>{1=>["length must be 1"]}}}
```


misdoro:
Thank you @timriley :pray:

PR #516, opened 2026-02-19
Fix bug with JSON Schema output for arrays with union items

### LINKED ISSUE #515: Incorrect JSON Schema generation of composed schemas in array items
## Describe the bug

When composing schemas as part of an array item (specifically in my case, a union between two schemas) the JSON Schema output is incorrect, producing something equivalent to `Array.of(Schema1) | Array.of(Schema2)` rather than `Array.of(Schema1 | Schema2)`.

## To Reproduce

```ruby
schema1 = Dry::Schema.JSON { required(:bar).value(:string) }
schema2 = Dry::Schema.JSON { required(:baz).value(:integer) }
Dry::Schema.JSON { required(:foo).value(:array).each { schema1 | schema2 } }.json_schema
```

## Expected behavior

Expected JSON Schema output:

```json
{
  "type": "object",
  "properties": {
    "foo": {
      "type": "array",
      "items": {
        "anyOf": [
          {
            "type": "object",
            "properties": {
              "bar": {
                "type": "string"
              }
            },
            "required": [
              "bar"
            ]
          },
          {
            "type": "object",
            "properties": {
              "baz": {
                "type": "integer"
              }
            },
            "required": [
              "baz"
            ]
          }
        ]
      }
    }
  },
  "required": [
    "foo"
  ]
}
```

Actual JSON Schema output:

```json
{
  "type": "object",
  "properties": {
    "foo": {
      "type": "array",
      "anyOf": [
        {
          "type": "array",
          "items": {
            "type": "object",
            "properties": {
              "bar": {
                "type": "string"
              }
            },
            "required": [
              "bar"
            ]
          }
        },
        {
          "type": "array",
          "items": {
            "type": "object",
            "properties": {
              "baz": {
                "type": "integer"
              }
            },
            "required": [
              "baz"
            ]
          }
        }
      ]
    }
  },
  "required": [
    "foo"
  ]
}
```
## My environment

- Ruby version: 4.0.1
- OS: macOS 15.6

adamransom:
Fixes #515 

When an array item contains schema combinations (via something like unions) the JSON Schema output is incorrect, producing the equivalent to

    Array.of(Schema1) | Array.of(Schema2)

rather than the correct

    Array.of(Schema1 | Schema2)

I'm not entirely sure if this is the best way to fix it, as I'm fairly unfamiliar with the codebase, but it seems a simple enough change and keeps the tests passing. Happy to be shown the correct way!

### DIFF
--- a/lib/dry/schema/extensions/json_schema/schema_compiler.rb
+++ b/lib/dry/schema/extensions/json_schema/schema_compiler.rb
@@ -133,13 +133,19 @@ module Dry
 
         # @api private
         def visit_or(node, opts = EMPTY_HASH)
+          any_of = []
           node.each do |child|
             c = self.class.new(loose: loose?)
             c.keys.update(subschema: {})
-            c.visit(child, opts.merge(key: :subschema))
-
-            any_of = (keys[opts[:key]][:anyOf] ||= [])
+            c.visit(child, opts.except(:member).merge(key: :subschema))
             any_of << c.keys[:subschema]
+          end
+
+          target = keys[opts[:key]]
+          if target[:type]&.include?("array")
+            target[:items] = {anyOf: any_of}
+          else
+            target[:anyOf] = any_of
           end
         end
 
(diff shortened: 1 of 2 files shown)


timriley:
Thanks again @adamransom! 😍

PR #518, opened 2026-06-26
Propagate dry-types meta and defaults into json_schema output; add .documentation macro

bogdan:
## Background

This PR addresses two recurring forum requests:
- [dry-schema key annotations](https://discourse.dry-rb.org/t/dry-schema-key-annotations/1544)
- [Getting dry-types metadata into dry-schema's JSON Schema output](https://discourse.dry-rb.org/t/getting-dry-types-metadata-into-dry-schemas-json-schema-output/1439)

## Summary

The `json_schema` extension currently has no way to attach descriptive metadata (description, title, examples, etc.) to individual fields without defining a separate named `dry-types` constant with `.meta(json_schema: {...})`. This PR makes that metadata first-class in two ways.

### 1. Propagate `dry-types` `.meta(json_schema: {...})` into JSON Schema output

When a type has `:json_schema` metadata attached via dry-types `.meta(...)`, it is now merged into the corresponding property in the JSON Schema output:

```ruby
Email = Dry::Types["string"].meta(json_schema: { description: "User email" })

schema = Dry::Schema.JSON do
  required(:email).filled(Email)
end

schema.json_schema
# => { ..., properties: { email: { type: "string", minLength: 1, description: "User email" } } }
```

This works for flat keys, nested hashes, and array-of-hash members. The `SchemaCompiler` receives the processor's `type_schema` and propagates it into child compilers for nested schemas.

### 2. Auto-populate `default` from dry-types default values

If a type has a runtime default set via dry-types, it is automatically reflected as `default` in the JSON Schema output by walking the type chain for a `Dry::Types::Default` node. This keeps the JSON Schema annotation in sync with the actual runtime behaviour without any extra effort from the user.

```ruby
required(:role).value(Dry::Types["string"].default("user".freeze))
# => { type: "string", default: "user" }
```

### 3. New `.documentation(...)` macro method

A `.documentation` method is added to `Macros::Core` by the `json_schema` extension, enabling inline annotation via method chaining without needing a named type constant:

```ruby
Dry::Schema.JSON do
  required(:email).filled(:string).documentation(description: "User email", example: "a@b.com")
  optional(:age).filled(:integer).documentation(description: "Age in years", deprecated: true)
  required(:address).hash { required(:street).filled(:string) }.documentation(title: "Address")
end
```

Accepted keys are the JSON Schema annotation keywords: `title`, `description`, `examples`, `example`, `deprecated`. The `deprecated` key defaults to `false` and is omitted from the output unless `true`. Passing `default:` raises `ArgumentError` — use dry-schema's native default instead, which is picked up automatically (see above). Calling `.documentation` before `.filled`/`.value` raises `Dry::Schema::InvalidSchemaError`.

## Test plan

- All existing `json_schema` extension specs pass unchanged
- New specs cover: meta propagation for flat/nested/array schemas, auto-populated defaults, all `.documentation` keys and their output, wrong-order guard, `deprecated` true/false behaviour, and hash key annotation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Break Versioning](https://www.taoensso.com/break-ve
 
 ### Added
 
+- `json_schema` extension now propagates `:json_schema` metadata from `dry-types` `.meta(json_schema: {...})` into JSON Schema property output (@bogdan)
+- `json_schema` extension now auto-populates `default` in JSON Schema output from dry-types default values (@bogdan)
+- Added `.documentation(title:, description:, examples:, example:, deprecated:)` macro method for attaching JSON Schema annotation metadata inline when using the `json_schema` extension (@bogdan)
+
 ### Changed
 
 ### Deprecated
--- a/lib/dry/schema/extensions/json_schema.rb
+++ b/lib/dry/schema/extensions/json_schema.rb
@@ -17,13 +17,39 @@ module Dry
         #
         # @api public
         def json_schema(loose: false)
-          compiler = SchemaCompiler.new(root: true, loose: loose)
+          compiler = SchemaCompiler.new(root: true, loose: loose, type_schema: type_schema)
           compiler.call(to_ast)
           compiler.to_hash
+        end
+      end
+
+      module MacroMethods
+        # Attach JSON Schema documentation metadata to this key
+        #
+        # @example
+        #   required(:email).filled(:string).documentation(description: "User email", example: "a@b.com")
+        #
+        # @return [self]
+        #
+        # @api public
+        def documentation(title: nil, description: nil, examples: nil, example: nil, deprecated: false)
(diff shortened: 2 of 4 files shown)


timriley:
Thank you for this, @bogdan! I just wanted to let you know that I'll be able to look at this in a week or so, I'm just focused on getting a Hanami release done in the meantime :)

bogdan:
Take your time: patches are cheap these days and human review is scarce. I'll remind you "in a week or so".

flash-gordon:
Default value can be any Ruby object, json schema cannot survive this. It only makes sense for JSON-compatible types. For instance, there's no type for date and it's easy to imagine a schema with a default date value.

bogdan:
@flash-gordon I implemented the default support for most common ruby types. That required some recursive mumba-jumba though. Take a look if it is something that can solve the problem.

bogdan:
@flash-gordon any feedback here?

gauravs:
Would be useful to get this merged

